PR #67, opened 2023-10-29
Enable storage transactions to match cronicle default

stennie:
Per [Storage Transactions](https://github.com/jhuckaby/Cronicle/wiki/Troubleshooting#storage-transactions) on the Cronicle wiki:
> It is highly recommended that you enable storage transactions. This protects against database corruption for events like backend storage errors, crashes and sudden power loss. Cronicle ships with transactions enabled in the default config, but if you have an older version you may need to manually enable them. 

This is also the default in upstream: https://github.com/jhuckaby/Cronicle/blob/master/sample_conf/config.json#L78

### DIFF
--- a/sample_conf/config.json
+++ b/sample_conf/config.json
@@ -86,8 +86,8 @@
 		"concurrency": 4,
 		"log_event_types": { "get": 1, "put": 1, "head": 1, "delete": 1, "expire_set": 1 },
 
-		"transactions": false,
-		"trans_auto_recover": false,
+		"transactions": true,
+		"trans_auto_recover": true,
 		
 		"Filesystem": {
 			"base_dir": "data",


mikeTWC1984:
So this was deliberate. Initially transaction didn't work on Windows setups. It was fixed later, but then I realized it's not that useful. It double pressure on filesystem, but that issue with was reported once in the last several years, and was related to some local setup of minio s3 storage. I tried to reproduce it on linux, crashing heavily loaded cronicle instance about 500 times, and never got partial json. In real prod environment the chance of this to happen is close to 0. So I'd only use it storage using some laggy driver (S3 or my sftp plugin). From the other side I have LMDB storage plugin which is atomic/poweroff safe, so I'd just use that instead of transacions.

stennie:
Thanks for the context @mikeTWC1984 !

Issue #41, opened 2022-10-07
Is SSO integration available?


audetto:
I mean something where I can delegate to a company service the user authentication.

1) is token present in cookies?
2) if not redirect to sso url (and go back to 1)
3) check with external system if token is valid
4) if yes, user is authenticated

mikeTWC1984:
No, neither original cronicle or this fork provide SSO.  I added a simple LDAP/AD for external authentication though.  Are you using some cloud auth provider?

audetto:
Ok.
No, on my organisation I can delegate all of this to an internal system.

If the mainpage redirects to

``http://internal.sso.provider/?returnUrl=http://address.of.cronicle``

then, when it comes back to `http://address.of.cronicle` there will be a cookie with a magic token.
At this point cronicle could verify the token and it is all done without anyone typing any password.

If this is too complicated, would it be possible to outsource to a new js module (that I can write for myself) the verification of the password?
Rather than cronicle having its own set of passwords, I could implement a function that checks if the password is correct, again delegating it to the company internal authentication provider.

mikeTWC1984:
I see. I think SSO is not that trivial to implement (in general). Might be an overkill for such indie tool. But if I get a chance, I'll definitely look into this.

If you just want to do custom password checking, you can take a look at my user.js implementation. It's pretty much dual mode, if you indicate user as external auth would go in custom route:
https://github.com/cronicle-edge/cronicle-edge/blob/main/lib/user.js#L210
You just need to replace do_ad_auth with your custom code. I also have some internal plumbing for external logins. Just keep in mind - you still need to register user and his permissions within cronicle  (check external box when creating user)

audetto:
>I see. I think SSO is not that trivial to implement (in general).

I would like to disagree with this statement: cronicle must already be using cookies to decide if a user is autheticated or it should be redirected to the (cronicle) login page.

If I could inject a function in that process which is allowed to inspect the coockies, I could check if the company SSO cookie is present (and valid) and in that case return the (trusted) name of the user (which of course must be known to cronicle).

If handling the redirection on a failure is too complicated, then it could continue as normal to the password page.

I would then bookmark a different URL for Cronicle, not `http://cronicle.company.com` but `http://sso.service/returUrl=http://cronicle.company.com` which means that when the code gets to the above function the token is always there.

[...]

PR #93, opened 2024-03-31
Add github oauth for user login

kc1r74p:
- Introduces Github OAuth Apps as an alternate auth flow -> getting away from username + password usage
- Partially integrated `auth_provider` which might enable further and cleaner ways of Auth in parallel

#### If a Github client is configured the Login mask will show an additional button:
grafik

Clicking that will redirect to the Configured Github instance to Grant the OAuth scope
Then Github will redirect back to Cronicle which will handle the `code`:
grafik

Afterwards a check is done in case `github_org` is configured, this will prevent any Github user gaining access and limiting the default access to members of configured organization, this could be extended to Teams or Users, for now org will already be a good improvement.

After login the user will see the default Cronicle Home page while automatically having his Github identity:
grafik

An `admin` user needs then to promote the new user in case it should also receive admin powers:
grafik

There are some configs which should suffice for custom Github Enterprise setups, in the future full CA and Cert handling would be cleaner and recommended:
grafik

Known issues / improvements left:
- Wanted to use octokit, but somehow had issues with bundling -> not nice for now
- Tokens are not revalidated ! -> Admins should disable Accounts in Cronicle once unused -> TODO ressume session/save token
- Allowing to use insecure instead of providing additional CA certs
- Also todo, check OAuth App for Server side user revocations -> token validity
- Changed External Auth, as it is partially reused to toggle UIs, not sure how clean if in use with AD / Github
- whole code base should be refactored/prettified/module normalized/typescripted :)

 
Feel free to raise some improvements, this should be a MVP but definitely could use some further enhancements as mentioned above.


mikeTWC1984:
Thanks for the PR, I was looking to add something like this but never got too far. I'll review. I was planning to get some better options for LDAP on the next major release, that could be nice addition.

danielangeloni:
with some minor refactoring to remove hardcode to github callbacks etc it could support any OAuth provider (i.e. Authentik or Authelia, Auth0 etc) 👀

spali:
seen it to late, but may my comment in https://github.com/cronicle-edge/cronicle-edge/issues/41#issuecomment-2488723327 help here to have a "plugin" api which would support and solve also the proxy case and with that almost all other authentication methods as the proxy is doing the heavy lifting of the auth.

PR #110, opened 2024-04-30
Update SQL.js to fix head operations for pg engine.

rrindels:
Postgres requires the "length" function vs "len" function as some SQL engines us, this patch would add the correct wrapping in the getBlobSizefn

### DIFF
--- a/engines/SQL.js
+++ b/engines/SQL.js
@@ -87,7 +87,7 @@ module.exports = class SQLEngine extends Component {
         let UPDATED = 'updated'
 
 	// Postgres requires quoted "V" field since the schema was created with the quoted "V" field. 
-	 T
+	 
         if (this.client === 'pg'){
 		this.getBlobSizefn = 'length("V")'
 	}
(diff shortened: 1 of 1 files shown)


mikeTWC1984:
Actually "lenght(V)" is a default value for  getBlobSizefn. It's only altered for MSSQL. Are you getting any error message referering to using "len"? 
https://github.com/cronicle-edge/cronicle-edge/blob/b89b35f7d2cedb9f211cbe5878002b1c5ee7dca1/engines/SQL.js#L45

rrindels:
Sorry i tried len, that was my bad, the fix was enclosing the V in quotes.

mikeTWC1984:
OK. good catch. Just one more thing - any reason you use escaped " ? Why no just  put statement in single quotes like:
```this.getBlobSizefn = 'length("V")'```

rrindels:
Because then I know its escaped  ;) no real preference just a style thing.

mikeTWC1984:
OK. I guess update your comment (since it's casing issue, not function), I'll merge

PR #121, opened 2024-06-05
Prevent params from logging to stdout

kc1r74p:
Similiar to https://github.com/cronicle-edge/cronicle-edge/pull/96 this change will hide potential Auth headers from the stdout log.

Without it, successful or failing jobs can easily leak Auth Headers to stdout which often is scraped with ELK / Loki in a k8s environment.

Simply added whole params object to exclude list, else it might be a bit ugly now using http-plugin logic here to search for the Auth header specifically. IMO `event_title` should be enough and details can still be viewed in real job run.

Prev:
grafik

With change:
grafik

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -285,8 +285,8 @@ module.exports = Class.create({
 
 	safeJobLog: function(job) { // print less verbose, more readable job data on logging
 		if(!job) return ''
-		let excl = ["table", "secret", "env", "cat_secret", "plug_secret", "globalenv"]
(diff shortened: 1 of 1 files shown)


mikeTWC1984:
I think a better approach would be making this a config, without altering current behavior. I'd even add an option to show all props if needed. E.g.
```json
"safe_job_log": {
  "show_params":true,
  "show_all": false
}
```

kc1r74p:
@mikeTWC1984 got some time and updated PR, would appreciate another review.

PR #122, opened 2024-06-06
Changes to allow running behind reverse proxy with path prefix.

mjavah:
- Changes to support relative pathing.
- Added external_prefix setting.
- Adjust socket.io and API path when external_prefix is provided.
- Added simple docker compose with reverse proxy setup.
- Shell scripts now keep their EoLs!


mikeTWC1984:
Thanks for PR. I think that's something people request once in a while. I started reviewing it, but for the future, open separate PR for each item, and also open an issue for each item so we can brain storm.

Question - is that something well tested on your end?
Can you add here a brief documentation on how to use proxy  path config?

mikeTWC1984:
Actually there is a simple way to achieve renaming manager node, just need to use manager with reset option, I just need to do a little tweak, so it will continue to setup if reset fails

mjavah:
Thanks!
It has been "lightly" tested.
There was no testing done for workflows or external auth though.
I've done some more testing and found a few more places where the "href"s needed to be relative.
Should I add them to this PR our create another one with just the "external_prefix"  changes?

Yes, I can document the "external_prefix" setting.  Where should that go?  Not seeing a /docs folder here.

Will the manager reset option remove the need for these k8s changes?

mikeTWC1984:
OK, submit your new changes to this PR.  I see your "K8" part is also removing old servers, that --reset option doesn't do. I guess your approach make sense for containers, since you don't really want to keep old ones.
One more though I have - if you resetting servers, probably it's a good idea o check if there is already running instance, to prevent accidental high jacking. Perhaps can add checking of "state" on the storage. Say if last change is newer than 1 minute that operation should fail.

mikeTWC1984:
TO clarify, if you set external_prefix  = cron, what should happen? If I set it, UI crashes, either on http://localhost:3012 and http://localhost:3012/cron. I was also trying to set nginx reverse proxy, but what I see in the errors is incorrectly concatenated ws uri (something like localhostcron).

mjavah:
The external prefix should be "/cron/".  Where would I document this?  In the Readme?

Issue #141, opened 2024-09-12
LDAP authentication issues


rgcouto:
Hello all,

I'm trying out this fork to check if I can successfully manage to login using LDAP (via JumpCloud), but this has been proven to be difficult to achieve, but maybe there's something wrong on my side that I'm simply not seeing.

I've currently even changed the `lib/user.js` file to hardcode the `opts` variable like described below:
```js
var opts = { 
	url: 'ldaps://ldap.jumpcloud.com:636',
	baseDN: 'uid=MY_USER,ou=Users,o=ORGANIZATION_ID,dc=jumpcloud,dc=com',
	password: 'MY_PASSWORD',
	tlsOptions: { 
		"rejectUnauthorized": false
	} 
}
```

I've also tried to login with the concatenation of the `user` variable with the `ad_domain` variable (which happens a couple lines below) disabled, but no luck either.

Checking out the logs, this is the only thing I get:
`[1726153623.194][2024-09-12 15:07:03][cronicle-edge-cronicle-edge-platform][47][User][debug][3][LDAP login failed][{"lde_message":"Invalid ldap packet","lde_dn":null}]`

What I know/tried to do:
- I've tried to look into the package code and can't find what might be wrong, but since the package is not being maintained anymore I wasn't sure if I got any reply there;
- Configuration strings from the LDAP server are correct, I've based my configuration from this URL from [JumpCloud](https://jumpcloud.com/support/use-cloud-ldap);
- I have confirmed that I have access to the LDAP server;
- I've tried to use `ldap://...:389`;
- I've also tried using the latest version of the `activedirectory2` package, version `2.2.0`.

Is there anyone who have had the same issue? If so, is there a way around this?

Thanks in advance.

mikeTWC1984:
That's right. I don't think at current configuration you can make it work with LDAP other then AD version of it.
Current module just verifying user/password using simple binding. In order to make this binding, you need to use full DN, which in AD can be replaced with SamAccountName. The latter can be used as username too. Technically you can use full DN as username, but it will have some special characters that not allowed. In general, LDAP auth needs some svc account to lookup user by mail/CN/etc, then perform a binding. Actually it was the next thing on my plate to add, so I'll try get it some time soon.

mikeTWC1984:
[...]

PR #142, opened 2024-09-18
improvement: Adds support for working with LDAP

rgcouto:
Following issue #141  the present PR aims to integrate LDAP support on cronicle-edge rather than only the AD version of it. 

The lib added to the tool was https://www.npmjs.com/package/ldapts.

To enable this feature set the config keys:
- `ldap_url`: The URL to your LDAP provider;
- `ldap_bind_dn`: The name (DN) of the directory object that the client wishes to bind as or the SASL mechanism (PLAIN, EXTERNAL);
- `ldap_opts`: The [TLS connect options](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback).

After some discussion in the beforementioned issue, the current PR has been reviewed and updated.

This solution is currently working for us.

If any other changes need to be made, please let me know.


rgcouto:
The new version of Cronicle Edge already addresses the proposed changes of this PR.

Issue #135, opened 2024-08-16
Feature request - enable/disable events only


Shto1337:
Hello,

We have cases in which some teams of developers only need to temporarily enable/disable events from the GUI(for example special sale events). Thus, add one more type of permission to enable/disable events(limited of course already to a group of servers/categories) but restrict everything else.
-> I know this is a part of "Edit events", but want to make sure they don't modify anything else(schedule/script/etc)

Would it be possible to add such a feature?

Shto1337:
I manage to do this, by editing the code(latest v1.9.5), with help of github copilot:

Added in config.json in client -> privilege_list:

{ "id": "disable_enable_events", "title": "Enable/Disable Events" }

And modified cronicle.js

```git diff
diff --git a/bin/cronicle.js b/bin/cronicle.js
index 3b1fe4f..cad8fbe 100755
--- a/bin/cronicle.js
+++ b/bin/cronicle.js
@@ -76629,6 +76629,8 @@ var require_event = __commonJS({
           });
         });
       },
+
+//##manual changes start here
       api_update_event: function(args, callback2) {
         var self2 = this;
         var params = args.params;
@@ -76645,22 +76647,74 @@ var require_event = __commonJS({
             return self2.doError("session", err.message, callback2);
           if (!self2.requireValidUser(session2, user, callback2))
             return;
-          if (!self2.requirePrivilege(user, "edit_events", callback2))
-            return;
-          if (params.abort_jobs && !self2.requirePrivilege(user, "abort_events", callback2))
-            return;
-          args.user = user;
-          args.session = session2;
-          self2.storage.listFind("global/schedule", { id: params.id }, async function(err2, event) {
-            if (err2 || !event) {
-              return self2.doError("event", "Failed to locate event: " + params.id, callback2);
+
+        // Check for both disable_enable_events and edit_events permissions
+        const canDisableEnableEvents = self2.requirePrivilege(user, "disable_enable_events", callback2, true);
+        const canEditEvents = self2.requirePrivilege(user, "edit_events", callback2, true);
+
+        if (!canDisableEnableEvents && !canEditEvents)
+          return self2.doError("event", "User does not have the required account privileges to perform this action", callback2);
+
+        if (params.abort_jobs && !self2.requirePrivilege(user, "abort_events", callback2))
+          return;
+
+        args.user = user;
+        args.session = session2;
+        self2.storage.listFind("global/schedule", { id: params.id }, async function(err2, event) {
[...]

PR #161, opened 2024-12-07
Feature request - enable/disable events only #135

Shto1337:


mikeTWC1984:
Hey, I noticed that new toggle API skips category/server group checking. This is needed in case user is limited by any of that. So we should add it back. 
From the other side, perhaps it would be easier just piggy back on existing update API. Perhaps just use query parameters to see what user want to do with it. E.g. check if query contains parameter like "enable", and if so just do custom permission check and  only use enabled property from submitted event
Query parameters can be checked via args.query object

Shto1337:
>> Hey, I noticed that new toggle API skips category/server group checking. This is needed in case user is limited by any of that. So we should add it back.

Solved and tested, reused the same check from previous function. Good catch :)

>> From the other side, perhaps it would be easier just piggy back on existing update API. Perhaps just use query parameters to see what user want to do with it. E.g. check if query contains parameter like "enable", and if so just do custom permission check and only use enabled property from submitted event
Query parameters can be checked via args.query object

IMHO it would not. If I understood correctly, while it sounds clever, I'm worried it could make things more complicated than they need to be because the the api_update_event function has a lot going on already with all the checks and updates it needs to handle. Adding extra logic for toggling could make it a bit messy and harder to manage, especially permission wise.
Lastly, my JS expertise is not that advanced and I would prefer to have it separate as a function, because its essentially a different permission type and easier code to read

mikeTWC1984:
OK.  In update api original code is also pulling category info, I guess just to make sure category still exists. Can you add same callback in you new api too? Like:
```js
				self.storage.listFind('global/categories', { id: params.category || event.category }, function (err, item) {
					if (err || !item) {
						return self.doError('event', "Failed to update event: Category not found: " + params.category, callback);
					}
```
So here is the callback sequence original code  is using:
load session => find event => find cat => update event

 Cat checking might be redundant, but perhaps not. Will be just easier to follow that pattern, than analyzing if it's needed.

Shto1337:
Ok, added

mikeTWC1984:
OK, I guess the last thing need to be reviewed - your api only works with "toggle event" permission. It should also work "Edit Event" permission

mikeTWC1984:
I mean if user have edit permission already, he still should be able to use that API. Currently "editors" need that "toggle" as addition.

Shto1337:
Oh, okay.
Fixed.
Just a note, i was not able to reuse requirePrivilege and used hasPrivilege because even though permissions were okay and the event was enabled/disabled, it returned visual error, and I had to check two permissions. Hope its ok.

Shto1337:
Mkay, 10x, changed.

mikeTWC1984:
Haha, this line makes admin whoever is editor or toggler:
```js
if (user.privileges.edit_events || user.privileges.disable_enable_events) return true;
```
It really needs to be 
```js
user.privileges.edit_events && priv_id === 'disable_enable_events') return true
```

Then in API you don't have to check  both edit and toggle, just toggle

Shto1337:
Huh, I did not expect that :).
See now.

mikeTWC1984:
Looks good. Thanks!

PR #175, opened 2025-01-21
Add volume management to Docker plugin

MaxwellJK:
- Mounting host path
 - Creating and mounting named volumes
 - Checkbox to control whether delete or not named volumes when the container is removed


mikeTWC1984:
OK, looks good.
Few things:
Make this parameter as "Text Field", not "Text Box". As a text box it look a bit busy, I guess not much users will use mounts or multiple mounts. For multiple mounts you can use comma separation, it's common pattern on this fork.

Other thing - keep default value blank. You can describe sample bindings in the script as commented lines
image

MaxwellJK:
hey @MaxwellJK thanks for your feedback - changes have been implemented.
I quickly tested and it seems to work as expected

steampoweredtaco:
Thanks for this.  I tested this on my existing instance. After updating the container via this branch, The new options did not appear in the existing Docker plugin. 

I had to basically grab the json for the plugin in the `sample_conf/setup.json` and import it as a new plugin and then I could see test and see mounting a local volume worked with the new plugin instance (nice).

I am new to cronicle, and not sure the proper update procedure when default plugins change.

Is there something missing here that should migrate the existing Docker plugin interface the first time after a plugin is updated?  Was there something I was suppose to do instead?

mikeTWC1984:
@steampoweredtaco  When you upgrade cronicle-edge, either with re-bundling or docker image upgrade, plugin binary will be updated to. However, plugin parameters are stored in you storage, so it's not get updated. In case new parameter added to plugin, you need to manually add this parameter. Or you can also import/export plugin in json format (on admin/plugins there is a button for this)
@MaxwellJK  those changes are good. I'll add them on the next release

PR #182, opened 2025-02-11
Adding Env Vars for setting and using Postgres DB as Storage

MaxwellJK:
storage-cli.json: adding a section to overwrite storage config if postgres variable is defined

### DIFF
--- a/bin/manager
+++ b/bin/manager
@@ -18,6 +18,7 @@ while (( "$#" )); do
         --key ) shift; key=$1 ;;
         --storage ) shift; storage=$1 ;;
         --sqlite ) shift; sqlite=$1 ;;
+        --sqlstring ) shift; sqlstring=$1 ;;
         --reset ) shift; reset=1 ;;
         --cluster ) shift; cluster=$1 ;;
         --debug ) shift; debug=1 ;;
@@ -35,6 +36,11 @@ fi
 if [[ $sqlite ]]; then
   export CRONICLE_sqlite="$(realpath $sqlite)"
   echo "Using custom sqlite db as storage: $sqlite"
+fi
+
+if [[ $sqlstring ]]; then
+  export CRONICLE_SQLSTRING="$(realpath $sqlstring)"
+  echo "Using custom db as storage: $sqlstring"
 fi
 
 if [[ $storage ]]; then
--- a/bin/manager
+++ b/bin/manager
@@ -39,8 +39,8 @@ if [[ $sqlite ]]; then
 fi
 
 if [[ $sqlstring ]]; then
-  export CRONICLE_SQLSTRING="$(realpath $sqlstring)"
-  echo "Using custom db as storage: $sqlstring"
+  export CRONICLE_SQLSTRING="$($sqlstring)"
+  echo "Using custom db as storage"
 fi
 
 if [[ $storage ]]; then
--- a/bin/manager.bat
+++ b/bin/manager.bat
@@ -43,6 +43,15 @@ if /I "%1"=="--port" (
     echo Using sqlite as storage: %~f2
     shift
     shift
+) else if /I "%1"=="--sqlstring" (
+    if "%2"=="" (
+      echo Connection string is not specified
+      exit
+    )    
+    set CRONICLE_SQLSTRING=%~f2
+    echo Using custom db as storage: %~f2
+    shift
+    shift
 ) else if /I "%1"=="--cluster" (
     if "%2"=="" (
       echo Missing cluster value. Specify comma-separatd hostnames
--- a/bin/manager.bat
+++ b/bin/manager.bat
@@ -48,8 +48,8 @@ if /I "%1"=="--port" (
       echo Connection string is not specified
       exit
     )    
-    set CRONICLE_SQLSTRING=%~f2
-    echo Using custom db as storage: %~f2
+    set CRONICLE_SQLSTRING=%~2
+    echo Using custom db as storage
     shift
     shift
 ) else if /I "%1"=="--cluster" (
--- a/bin/storage-cli.js
+++ b/bin/storage-cli.js
@@ -71,7 +71,7 @@ if(process.env['CRONICLE_sql_connection_string']) {
                 "host": cs['hostname'],
                 "port": cs['port'],
 				"user": cs['username'],
-				"password": cs['password'],
+				"password": decodeURI(cs['password']),
 				"database": cs['pathname'].slice(1)
 			}
 		}
--- a/bin/storage-cli.js
+++ b/bin/storage-cli.js
@@ -65,7 +65,7 @@ else if(process.env['CRONICLE_SQLSTRING']) {
                 "host": cs['hostname'],
                 "port": parseInt(cs['port']),
 				"user": cs['username'],
-				"password": decodeUriComponent(cs['password']),
+				"password": decodeURIComponent(cs['password']),
 				"database": cs['pathname'].slice(1)
 			}
 		}
--- a/bin/storage-cli.js
+++ b/bin/storage-cli.js
@@ -63,7 +63,7 @@ else if(process.env['CRONICLE_SQLSTRING']) {
 			"table": cs['searchParams'].get('table') || 'cronicle',
 			"connection": {
                 "host": cs['hostname'],
-                "port": cs['port'],
+                "port": parseInt(cs['port']),
 				"user": cs['username'],
 				"password": decodeUriComponent(cs['password']),
 				"database": cs['pathname'].slice(1)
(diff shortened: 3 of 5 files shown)


mikeTWC1984:
Looks good to me, perhaps can use as template for other DBs later. 
Did you test it on your end?

mikeTWC1984:
Btw, one of the option of thinking is to pass DB connection via URI. 
E.g. CRONICLE_SQLHOST = pgsql://user:password@test.com:1521
That would be kinda generic for most DBs and easy to parse. If you feel enthusiastic you can add this option too. 

image

MaxwellJK:
Hey @mikeTWC1984 
yes my version currently runs using this setup and it's storing the schedules on my postgres.
I am not sure i understand your second message - a bit more info pls? I might feel enthusiastic since I am home alone these days.

MaxwellJK:
ohh you mean defining a variable where users can provide a db string and URI converts it into params?

mikeTWC1984:
Yeah, so instead of setting up variable for each component (port/host/etc), user might set just 1 variable that would be URI for the DB. Connection info typically has host/port/user/password, which are URI components. Driver name could be used in placed of protocol. Some DB might require some extra params, but likely can be added to URI as query string (e.g. URI?key=value&key2=value). This way it also can be generic, so no need to specify different variables for Postrges/Mysql/etc

MaxwellJK:
Done - but not tested it.
If i find time, i will rebuild my cronicle at some point between tomorrow and monday to test it (although I am fairly confident it will work).
I didn't remove the variables for postgres yet since i didn't know if you wanted to manage the connections to db via connection string only or if you wanted both options. let me know and i can easily remove it in case.

I am not sure how to manage the extra params because I don't know where those would go in the Storage JSON configuration.

mikeTWC1984:
OK, few extra things. 

1) You can pass table name via query parameter, e.g.
```
pgsql://test:pass@some.db.company.com:1521?table=cronicle
```
On the parsed URL it will be stored as searchParams property
Same way other parameters can be passed, but at this point just table should be enough

2)
When reading password - add extra step to "decode" it. Password can have characters like @ or / and those supposed to be escaped  (e.g. @ would be %40)

3) I'm fine if you keep both ways. Just test URL method using your postgres connection. I can test it later on some other DBs

MaxwellJK:
just tested with my postgres and it works pretty well - i span a secondary container on my laptop and connected to the same database used by cronicle on my server and it worked flawlessly.

I tested the code with this connection string `pg://username:password@server:5432/?table=cronicle` and it works.
I tired using `pgsql://...` and it doesn't work. This is because I am using `pg` from the URL as value for the client when creating the connection.

I guess options here are:
- change the code so `pgsql` is translated into `pg`
- change the code so it accepts `pgsql` as client
- just document it works in that way and if the client is not recognised, cronicle fails with an error
```
Error: knex: Unknown configuration option 'client' value pgsql. Note that it is case-sensitive, check documentation for supported values.
    at sX (/opt/cronicle/bin/engines/SQL.js:82:25795)
    at Ze (/opt/cronicle/bin/engines/SQL.js:82:26315)
    at module.exports.setup (/opt/cronicle/bin/engines/SQL.js:82:27467)
    at module.exports.startup (/opt/cronicle/bin/engines/SQL.js:82:27410)
    at __construct.startup (/opt/cronicle/bin/cronicle.js:56:4503)
    at /opt/cronicle/bin/cronicle.js:9:4519
    at /opt/cronicle/bin/cronicle.js:2:18884
    at replenish (/opt/cronicle/bin/cronicle.js:2:7520)
    at /opt/cronicle/bin/cronicle.js:2:7570
    at eachLimit$1 (/opt/cronicle/bin/cronicle.js:2:18994)
```

let me know if you have a different idea on this.

Thanks

mikeTWC1984:
Nah, client is just name of the driver used by underlying SQL library (knex)
Here is only values accepted at the moment:
sqlite3, mysql2, pg, mssql, oracledb, 

Check this page for more info
https://github.com/cronicle-edge/cronicle-edge/tree/main/engines#sql

So one extra thing you can do - check if protocol name from the URI matching one of those 5 values, if not - print friendly error message listing supported drivers. You can add alias too like you suggest, converting pgsql to pg (maybe also mysql => mysql2, oracle to oracledb)

MaxwellJK:
Checking the protocol and friendly error message implemented
It doesn't check for sqlite3 since it's using a different env var (and more in general the connection string for sqlite3 is not in a format URI can interpret - or at least i didn't find a way).

mikeTWC1984:
OK, looks good.

I did some testing, this approach work for all 4 drivers. Now few more items.

1) use decodeUriComponent vs decodeURI (latter doesn't seem to decode % escaped characters)
2)  I think sql_connection_string is a bit verbose. Should we just go with CRONICLE_SQLSTRING?
    I also like to add such parameters to manager command. E.g. if I pass manager --sqlstring, it should create that variable (see how --sqlite paramter works)
3)  looks like on storage-cli those configs (for sqlite, sqlstring and postgres) are in separate IFs, not in a single if/else block

4) You also need to remove env variables with sensitive info after cronicle starts. Check engine.js line 91. Add there variables that may contain password (sqlstring, pg_password)

MaxwellJK:
> use decodeUriComponent vs decodeURI (latter doesn't seem to decode % escaped characters)

 Done


> I think sql_connection_string is a bit verbose. Should we just go with CRONICLE_SQLSTRING?
> I also like to add such parameters to manager command. E.g. if I pass manager --sqlstring, it should create that variable (see how --sqlite paramter works)

Done Both

> looks like on storage-cli those configs (for sqlite, sqlstring and postgres) are in separate IFs, not in a single if/else block

Done

> You also need to remove env variables with sensitive info after cronicle starts. Check engine.js line 91. Add there variables that may contain password (sqlstring, pg_password)

Done

> I think you can also resolve table name without if/else, just tb = params.get('table') || 'cronicle'

Done

> Also add proto2driver map to resolve driver, so user can use more intuitive name (e.g. mysql vs mysql2) 
> `{'pg:':'pg', 'pgsql:':'pg', 'mysql:':'mysql2', 'mysql2:':'mysql2','oracle:':'oracledb', 'mssql:':'mssql'}`

Not done - this approach can open a pandora's box where we will have a lot of mappings to accommodate having a more intuitive name. Also, the connection string is something that you set up potentially once in a liferime. Do we really want to complicate the logic so much just to allow a few synonyms?

mikeTWC1984:
OK, good work will review this tomorrow.
Regarding last point - not sure what the concern is. Logic won't be more complicated, just a different approach of validation
```js
let map = {'pg:':'pg', 'pgsql:':'pg', 'mysql:':'mysql2', 'mysql2:':'mysql2','oracle:':'oracledb', 'mssql:':'mssql'}
if(!map[cs.protocol]) throw new Error(`Invalid Driver, use on of the following ${Object.keys(map).map(e=>e.slice(0,-1))}`)
```
I like alias since it's hard to remember actual driver name. Not a big deal for average user, but say in a month or so I'd want to test all 4 engines. I will likely not remember that mysql is using mysql2 driver, so using name like mysql or oracle is kinda intuitive.

MaxwellJK:
No concerns, I was just raising because right now we have only 4 databases but if in future we have 6 or 7 and start adding 2 or 3 aliases for all of them it might get out of control - the complexity is not in the code.
But I also get your point so no big deal to add it in

mikeTWC1984:
OK, getting close. few more tweaks needed:
1) On manager and manager.bat - sqlite parameter expected to a file, so it resolves absolute file path. For sqlstring it's not needed, just use $sqlitestring / %2 for value. (without realpath/ ~f). Also don't print that string, it may contain password.

2) When using cs['port']  wrap it in parseInt. I noticed mssql driver freaks out if it's string type.

3) in  decodeURIComponent    URI is upper case, not Uri

Regarding mapping protocol, I think we won't add any more SQL DBs other than big 4. The only other I know out of top of my head is IBM DB2 / it's derivatives, which I don't think are used as general purpose DBs.

MaxwellJK:
> On manager and manager.bat - sqlite parameter expectes file, so it resolves absolute file path. For sqlstring it's not needed, just use $sqlitestring / %2 for value. (without realpath/ ~f). Also don't print that string, it may contain password.

ah sorry about that, i blindly copied without checking what the code was doing...

> When using cs['port'] wrap it in parseInt. I noticed mssql driver freaks out if it's string type.

Done

> in decodeURIComponent URI is upper case, not Uri

Done

mikeTWC1984:
Hopefully the last thing - I think also a good idea to store db password in separate variable from SQLSTRING. E.g. just use default CRONICLE_Storage__SQL__connection__password variable, maybe a shortcut like CRONICLE_SQLPASSWORD. In this case password from connection string will be null, so we need to fall back to those password variables.
I'll do final test tomorrow

mikeTWC1984:
OK, I think I'm good to merge it. There a few minor tweaks I might need to apply later before release, but will do it myself

PR #244, opened 2026-08-06
fix(build): declare direct bundle dependencies

frankii91:
## What changed

- Declare `he`, `lodash`, and `pixl-cache` as direct runtime dependencies.
- Keep the root dependency list in `package-lock.json` in sync.

## Why

The bundle source imports these packages directly:

- `bin/workflow.js` imports `he`
- `lib/engine.js` and `lib/api/confkey.js` import `lodash`
- `engines/Filesystem.js` imports `pixl-cache`

They were only available through transitive dependency hoisting. With a strict dependency layout such as pnpm, `./bundle dist` failed with `Could not resolve` errors for all three modules. Declaring them directly makes the build deterministic instead of relying on an installer-specific tree layout.

## Impact

No application behavior or configuration changes. This only makes existing direct imports explicit and allows the bundler to resolve them reliably.

## Validation

- `corepack pnpm install --lockfile=false --ignore-scripts`
- `corepack pnpm test`: 54/54 tests passed, 357 assertions
- On FreeBSD 14.3, the bundle completed without the three module-resolution failures and a shell smoke job completed with exit code 0


mikeTWC1984:
OK, good point

PR #246, opened 2026-08-07
Add configurable OAuth-only login

frankii91:
## What changed

- `oauth.button_label` sets the OAuth/OIDC button text.
- `oauth.only` hides local login controls and rejects password login requests at `/api/user/login`.
- `oauth.auto_login` redirects unauthenticated visitors directly to OAuth/OIDC.

All options keep the current behavior by default.

## Why

OIDC deployments often need a provider-specific label and a way to enforce SSO without reverse-proxy or local frontend patches.

## Validation

- 55/55 unit tests passed (364 assertions).
- JavaScript syntax and sample config JSON validated.
- Login UI checked for default, custom label, OAuth-only, auto-login, and OAuth-disabled configurations.

Manual browser verification against a real provider is pending.

PR #247, opened 2026-08-07
Add configurable OAuth-only login

frankii91:
## Summary

- add a configurable OAuth login button label
- add an optional OAuth-only mode that hides local account actions and rejects password login at the API
- add optional automatic redirect to the configured OAuth provider
- document the new settings and include safe defaults in `sample_conf/config.json`

This extends the existing OAuth support documented in the [v1.11.1 release notes](https://github.com/cronicle-edge/cronicle-edge/releases/tag/v1.11.1). All new options default to disabled, so existing login behavior remains unchanged.

## Configuration

```json
"oauth": {
  "enabled": true,
  "button_label": "Login with Keycloak",
  "only": true,
  "auto_login": false
}
```

Operators should verify their OAuth provider flow before enabling `only` or `auto_login`, because those modes intentionally remove the interactive local-login fallback.

## Validation

- added coverage that OAuth-only mode rejects password login and that the new client configuration defaults are exposed correctly
- OAuth login and logout were manually tested end-to-end by a human operator on a physical FreeBSD host

The implementation uses the existing platform-independent JavaScript and configuration paths and contains no FreeBSD-specific code, so it should work on supported Linux deployments as well.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -103,6 +103,21 @@ Run your script in actual terminal emulator (using node-pty) in order to capture
 Can verify user via common oauth providers (github, google, authentik, microsof, keyclock, etc)
 Refer to [1.11.1 release notes](https://github.com/cronicle-edge/cronicle-edge/releases/tag/v1.11.1) for more details.
 
+The login page can be customized in the `oauth` section of `config.json`:
+
+```json
+"oauth": {
+  "enabled": true,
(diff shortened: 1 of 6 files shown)


frankii91:
Closing as an accidental duplicate of #246. Both pull requests point to the same commit and contain the same diff.

PR #248, opened 2026-08-07
feat(oidc): add secure RP-initiated and back-channel logout

frankii91:
## Summary

- add optional OpenID Connect RP-Initiated Logout;
- add signed OIDC Back-Channel Logout with `sid` and `sub` session revocation;
- verify ID Tokens and Logout Tokens with remote JWKS;
- protect session creation and resume from concurrent logout;
- redact OIDC secrets before request logging.

Both logout mechanisms remain disabled by default. OIDC still requires an existing local Cronicle account and does not change roles or privileges.

## Review updates

- moved OIDC internals into the `lib/oidc.js` mixin so `lib/user.js` follows the existing Cronicle structure;
- added short comments to the OIDC methods where the session and locking flow is not obvious;
- moved detailed configuration to `docs/oidc.md` and kept only a short overview in the README;
- separated `allow_http_localhost` from `insecure`: local HTTP can be explicitly allowed without disabling HTTPS certificate verification;
- aligned Logout Token validation with the final specification: `exp` is optional, but tokens without it still have a finite window based on `iat` and `max_token_age_seconds`;
- made OIDC session resume use the same revocation locks, so an in-flight resume cannot restore a logged-out session.

## Back-channel behavior

Endpoint:

`POST /api/user/oidc_backchannel_logout`

Cronicle validates the signature, algorithm, `iss`, `aud`, `iat`, `jti`, logout event, absence of `nonce`, and the required `sid` or `sub`. If `exp` is present, it is also validated.

A `sid` token removes the matching provider session. A `sub` token removes all indexed sessions for that provider subject. Replays and already removed sessions are handled idempotently.

Callback, resume, and back-channel revocation share ordered identity locks and persistent short-lived revocation markers. A delayed callback or resume therefore cannot recreate an invalidated session.

## RP-Initiated Logout

Local session state is removed before redirecting to the provider. The verified ID Token is kept encrypted in server-side storage and exposed only through a short-lived, encrypted, one-time redirect ticket. Access and refresh tokens are not persisted.

An OIDC session without a verified ID Token fails closed instead of silently falling back to `client_id`.

## Validation

- `git diff --check`;
- JavaScript syntax checks;
- clean trial merge with current `upstream/main`;
- PR branch: 89/89 tests, 364 assertions;
- integration branch with current FreeBSD changes: 90/90 tests, 369 assertions;
- focused OIDC suite on FreeBSD: 34/34 tests;
- production OIDC login and RP-Initiated Logout smoke test passed on FreeBSD.

## Documentation

- [OpenID Connect Back-Channel Logout 1.0 Final](https://openid.net/specs/openid-connect-backchannel-1_0-final.html)
- [OpenID Connect RP-Initiated Logout 1.0](https://openid.net/specs/openid-connect-rpinitiated-1_0.html)

The former dependency, #246, is now merged into `main`. This PR contains the three focused logout commits on top of it.


mikeTWC1984:
OK, sounds like a good stuff, it's just quite big change, will take me some time to review/understand. As far as I remember, I haven't implemented OIDC logout properly (just did local session invalidation), or something like that.  So it will be helpful

frankii91:
After a great deal of work, we now have complete and consistent logout
and session invalidation implemented and working correctly according to the
relevant specifications. It operates across multiple integrated systems,
including Cronicle Edge, Cloudflare Access, three PHP applications, and
soon a C# application, with more services to follow. This is an important
technical consideration when building a unified environment with SSO and a
consistent session lifecycle across all connected applications.

mikeTWC1984:
For "ouath" only scenario - what are your thoughts on locking your self in case when OIDC is not functioning? 

I'd also do some more manual testing using other mainstreams providers. I'd do authentik and/or authelia (I guess most common opensource) and  google or github (non opensource) . I recall I had to tune up login process when trying different providers. I have authentik in my dev, I'll check, if you have capability check google or github - would be nice

Regarding docs - it's a bit verbose for the readme page, I'd probably put technical details / detailed samples in a separate file like docs/oidc.md and put a link to it on Readme. I usually put detailed description in release notes

lib/api/admin.js
mikeTWC1984, line 1: That's OK for now - but for future - can you wrap such things in a separate commit, or even PR. I think your main PR comment doesn't mention this.  I generally agree with this change , I added that debug option since it was a PITA during testing bunch of different storage system / multiple instances.  I guess if I need it I can just do this temporarily during development.

mikeTWC1984:
One more thing - oidc.js seemed to be just a helper util. Then you also define bunch of helpers in user.js (I guess because you need access to storage and debug printing?). This kinda bloats user.js, and is not in line with "Cronicle" desing pattern. Consider convert this to "mixin" or "component". I guess first option is what would be in-line with cronicle idea. You can use engine.js as an example - it is loading several files (e.g. job.js) from lib folder. If you load it this way those classes will share "this" object, so you can refer to self.storage inside oidc.js, same way it is done in job.js for example
Generally all files in lib should be either "components" or "mixins"

mikeTWC1984:
so basically try to add __mixins property in user.js, then all methods you define in oidc.js should be available within user.js (and vice versa)

mikeTWC1984:
Also, there are bunch of new oidc functions, but I  don't see much comments, so it's hard to follow what those are doing. Please add comments

frankii91:
Yes, that makes sense. This is a different approach than mine. I prefer one
comprehensive commit/project to many small ones (for me it's easier to
physically test and achieve a realistic goal), and maybe that's the problem.

frankii91:
Thanks for the tip! I'm going on a longer vacation. If I can manage it (or
someone else can), I'll fix the whole PR. Just let me know the best way to
do it: one new fix, or split it into several smaller ones? I've only been
doing PRs for open source for a short time, so I have absolutely no
experience working with others, and I'm coming to the conclusion that if I
need something and need to do it for myself, it's worth pushing it into the
source so I don't have to maintain my own forks, but so that most of it
goes into the source.

frankii91:
> W przypadku scenariusza podstawowego na protokole „out-to-out” – co sądzisz o możliwości zablokowania się w przypadku, gdy OIDC nie działa?
> 
> Przeprowadziłbym też więcej testów ręcznych, zastępujących usługi innych niebezpiecznych dostawców. Wykorzystałbym Authentik i/lub Authelia (chyba najpopularniejszy open source) oraz Google lub GitHub (nie open source). Pamiętam, że musiałem dostroić proces logowania, różnych dostawców. Mam Authentik w moim środowisku deweloperskim, sprawdzę. Jeśli masz taką możliwość, sprawdź Google lub GitHub – niesamowitą miłość.
> 
> Jeśli chodzi o konstrukcję – jest trochę za długa jak na stronę readme. inspektor kryminalny szczegóły techniczne / szczegółowe dowody w pliku plik, taki jak docs/oidc.md, i odkryłbym do niego link w readme. zamieszczam szczegółowy opis w notatkach o wydaniu.

Yes, I am aware of the risk of being locked out if OIDC fails, and I am going in this direction consciously. I still believe it is a better solution than maintaining separate authentication mechanisms, separate accounts, and separate security policies in every service, not to mention the problems with user passwords, password rotation, and different authentication methods across multiple systems.

I prefer to have one well-protected "vault with the keys" and standardize on one or two strong authentication methods for myself and other users.

Of course, I still keep an emergency path. In the applications I maintain a simple configuration switch that allows me, in case of an OIDC failure, to disable OIDC and restore traditional login/password or Passkeys authentication.

In parallel, I also use a break-glass approach. There is a separate administrative account and authentication mechanism that is normally completely disabled and can only be enabled through a configuration change in an emergency.

However, my main priority is to restore the proper authentication environment quickly rather than operate for a long time using emergency mechanisms. In my case this environment includes privacyIDEA, Keycloak, LDAP, supporting Docker services, consistent automated backups of both data and configuration, Docker Swarm running inside the organization's local network rather than externally, and automated failover of the Internet connection.

I am also currently developing my own fork of the privacyIDEA mobile application, with the main focus on PUSH tokens. One of the key changes I am working on is much stronger binding between the token and the user's biometrics on the device.

The idea is that, from the moment the token is issued, its cryptographic key should be bound to strong biometric authentication. It should not be possible to fall back to a device PIN or password. A change to the enrolled biometric set on the device should invalidate the existing token and require a new enrollment.

I already have this implemented, but at the moment I have too few devices and testers to say with complete confidence that it behaves exactly as expected across all real-world Android and iOS scenarios. According to the platform documentation, the mechanisms should support this. But as usual, documentation and actual behavior across different devices are not always exactly the same.

So despite the additional dependency on a centralized authentication system, I still consider this a better approach than maintaining distributed and independent authentication in every individual service, provided that proper redundancy, backups, and break-glass mechanisms are in place.

mikeTWC1984:
OK, good. Keep same PR (no need to split), add comments for your oidc functions and try to turn oids.js to "mixin", or at least put all oidc logic internals in your oidc module. Generally I'd prefer 1 PR per feature, it's just easier to digest and do isolated testing, 
There is no rush,  take your time. I still want to understand better those changes, and test on different OIDC providers.

mikeTWC1984:
Trying to get it work with authentik. So far - login fails with 
[unexpected "iss" claim value]
Before it was complaining about not getting 200 for jwks url, but somehow I managed to fix it.

Can this cert/claim/whatever check be opt-out by setting? It should work for all providers, if not then there should be an option to opt it out. 

Also, there are 2 parameters checking/declining  for localhost (general and logout) . Why is it needed? I'd obviously would use http://localhost for testing, just not sure what is rationale to block it?

Here is the config I use for authentik, You can install it locally with docker compose. Or just make all those error you do very detailed. 

```json
	"oauth": {
		"enabled": true,
		"button_label": "SSO",
		"only": false,
		"auto_login": false,
		"insecure": false,
		"client_id": "****",
		"client_secret": "*****",
		"redirect_uri": "http://localhost:3034/api/user/callback",
		"authorize_url": "http://localhost:9000/application/o/authorize/",
		"token_url": "http://localhost:9000/application/o/token/",
		"user_url": "http://localhost:9000/application/o/userinfo/",
		"user_attribute": "email",
		"scope": "email openid profile",
		"params": {
			"access_type": "offline",
			"prompt": "consent"
		},
		"headers": {
			"accept": "application/json"
		},
		"issuer": "http://localhost:9000/application/o/cronicle",
		"jwks_url": "http://localhost:9000/application/o/cronicle/jwks/",
		"allowed_algs": ["RS256"],
		"clock_tolerance_seconds": 5,
		"jwks_cooldown_seconds": 30,
		"allow_http_localhost": true,
		"logout": {
			"enabled": true,
			"end_session_url": "http://localhost:9000/application/o/cronicle/end-session/",
			"post_logout_redirect_uri": "http://localhost:3034",
			"allow_http_localhost": true,
			"params": {}
		},
		"backchannel_logout": {
			"enabled": false,
			"max_token_age_seconds": 300,
			"max_token_size": 16384
		}
	},
```

mikeTWC1984:
Actually nvm, the issuer url was just missing / and I should use preferd_user instead of email in my setup

mikeTWC1984:
here is working example for reference

```json
	"oauth": {
		"enabled": true,
		"button_label": "SSO",
		"only": false,
		"auto_login": false,
		"insecure": false,
		"client_id": "****",
		"client_secret": "*****",
		"redirect_uri": "http://localhost:3034/api/user/callback",
		"authorize_url": "http://localhost:9000/application/o/authorize/",
		"token_url": "http://localhost:9000/application/o/token/",
		"user_url": "http://localhost:9000/application/o/userinfo/",
		"user_attribute": "preferred_username",
		"scope": "email openid profile",
		"params": {
			"access_type": "offline",
			"prompt": "consent"
		},
		"headers": {
			"accept": "application/json"
		},
		"issuer": "http://localhost:9000/application/o/cronicle/",
		"jwks_url": "http://localhost:9000/application/o/cronicle/jwks/",
		"allowed_algs": ["RS256"],
		"clock_tolerance_seconds": 5,
		"jwks_cooldown_seconds": 30,
		"allow_http_localhost": true,
		"logout": {
			"enabled": true,
			"end_session_url": "http://localhost:9000/application/o/cronicle/end-session/",
			"post_logout_redirect_uri": "http://localhost:3034",
			"allow_http_localhost": true,
			"params": {}
		},
		"backchannel_logout": {
			"enabled": false,
			"max_token_age_seconds": 300,
			"max_token_size": 16384
		}
	},
```

frankii91:
Thanks for the review. I kept this in the same PR and applied the requested cleanup:

- OIDC internals now live in the `lib/oidc.js` mixin;
- the less obvious session and locking methods have short comments;
- detailed configuration moved to `docs/oidc.md`;
- `allow_http_localhost` now works independently from `insecure`, matching your Authentik localhost setup;
- Logout Tokens without `exp` are accepted as required by the final specification, with a finite window based on `iat`;
- session resume now shares the logout locks and cannot restore a revoked session.

I left `admin.js` unchanged in this review update, as discussed.

Validation passed: 89/89 tests on the PR branch, 90/90 on the integration branch, and 34/34 focused OIDC tests on FreeBSD. Production OIDC login and RP-Initiated Logout also passed.

frankii91:
To clarify, I do not mean additional known defects in this PR. While comparing the same logout flow with other session implementations, I found a few recurring hardening areas:

- a delayed login callback can create or reactivate a session after back-channel logout when only the replay `jti` is stored and there is no durable `issuer + client_id + sid/sub` revocation marker;
- storing the replay `jti` before the revocation marker and session deletion can leave a crash window: a retry is rejected as replay while the session may still be active;
- local session revocation can depend on later provider, configuration or database steps instead of being guaranteed first or in a `finally` path;
- the boundary between an OIDC library and the host application may not clearly require session creation and revocation to share one lock or transaction.

This PR already closes the callback/resume race that is relevant to Cronicle. I left the broader findings out to keep the current change focused. Before proposing more code, I would first verify which points actually apply to Cronicle and reduce them to concrete failing tests.

Would you prefer a short issue with those verified cases and proposed tests first, or a separate small PR only for anything that can be demonstrated in Cronicle, so you can review and evaluate it in your own time?

frankii91:
Manual Google provider test completed on the exact PR head (`f4ad553`).

- Used a temporary Google web client with `http://localhost:3012/api/user/callback` and `openid email profile`.
- Two complete browser cycles passed: authorization, callback/code exchange, signed ID Token validation (issuer, audience, nonce and Google JWKS), UserInfo mapping, Cronicle session creation, and local logout.
- After the final logout, reloading Cronicle still showed the login page, so the deleted session was not resumed.
- The full local suite also passes: 89/89 tests and 364 assertions. Syntax checks for all changed JavaScript files and `git diff --check` pass.

Google supports OIDC authentication, but its current discovery document does not publish an `end_session_endpoint` or back-channel logout support. Therefore Google can validate provider compatibility, ID Token handling and local logout, but it cannot validate RP-Initiated Logout or Back-Channel Logout. Those features still need a provider such as Authentik or Keycloak.

I did not create a separate GitHub test app. Login with GitHub is OAuth 2.0 rather than end-user OIDC: it returns an access token and `/user` data, but no OIDC ID Token or standard logout endpoints. In Cronicle it would only exercise the generic authorization-code, UserInfo and local-logout path, without testing the OIDC logout feature, so it would add little after the Google test and the existing regression suite.

One thing I still do not understand is why the [Docker / build (pull_request)](https://github.com/cronicle-edge/cronicle-edge/actions/runs/31857498891/job/94944917997?pr=248) check is failing. I may be doing something wrong when preparing the PR. I have not tried to change it because I do not want to mix an unrelated CI change into this OIDC work.

frankii91:
I checked the Docker failure. It is not caused by this PR.

The workflow runs the DockerHub publishing step for `pull_request`, but DockerHub secrets are empty for fork PRs and the script only handles `push` and `release`. It therefore fails at `docker login` before the image is built.

Could you consider building without login/push on `pull_request`, and keeping DockerHub publishing only for `push` or `release`? Otherwise external PRs remain red even when their code and tests are fine.

mikeTWC1984:
OK, good. So the main thing I really wanted to check if that that new jwks/logout flow would not break current authorization flow. Which seemed to be the case, my old test config without jwks/logout url still works fine. I'll likely will merge it soon

Curios - what's wrong with using localhost? There are 2 allow_localhost flag + unsecure, I guess in most cases you need them all together

mikeTWC1984:
For those docker errors - you can ingore them for now, I don't use it for decision making. But I'll try to tweak actions file

mikeTWC1984:
OK, looks good now. I guess I'm ready to merge. Let's just clarify on those localhost configs. I'm ok to disallow localhost, but can we just use "insecure" to cover all 3 options (http + localhost for auth/logout)? Or keep all 3 but treat "insecure" as combo.

frankii91:
Thanks for asking. I think keeping the three settings independent is safer and more precise.

- `oauth.insecure` is a pre-existing setting. It only disables TLS certificate verification (`rejectUnauthorized`) for the token and UserInfo requests, so it is useful for HTTPS with a self-signed or otherwise untrusted certificate.
- `oauth.allow_http_localhost` permits cleartext HTTP only for loopback issuer and JWKS URLs.
- `oauth.logout.allow_http_localhost` separately permits cleartext HTTP only for loopback end-session and post-logout URLs.

They are not always used together. A self-signed HTTPS provider may need only `insecure`; a public HTTPS provider used by a local Cronicle instance may need only the logout localhost allowance; normal production needs none of them. Making `insecure` a combo would silently broaden an existing TLS-only setting to also permit cleartext URLs and logout behavior, so I would prefer to keep the explicit flags.

I agree that the name `insecure` is ambiguous, and your question shows why. Since it predates this PR and may already be used by existing configurations, renaming it here would break compatibility. After this is merged, it may be worth adding a clearer `skip_tls_verification` option, keeping `insecure` as a backward-compatible deprecated alias, and using the clearer name in the sample configuration and documentation.

frankii91:
A reverse proxy or tunnel is another practical reason to keep these controls separate. There are independent TLS legs: client to proxy, proxy to Cronicle, and Cronicle to the OIDC provider. A private or self-signed certificate between Cloudflare/cloudflared and the Cronicle origin is configured on the proxy/tunnel side and does not require `oauth.insecure`.

`oauth.insecure` is relevant only when Cronicle itself connects to the OIDC provider over HTTPS and cannot verify that provider certificate. If Cronicle reaches the provider through its normal public HTTPS hostname, all OIDC transport exceptions can remain disabled even when the proxy uses a private certificate toward the Cronicle origin. Local loopback HTTP is a different exception and is therefore controlled by the `allow_http_localhost` settings.

lib/oidc.js
```diff
@@ -0,0 +1,798 @@
...
+				'Cache-Control': 'no-store',
+				'Referrer-Policy': 'no-referrer'
+			});
+			args.response.end();
+			return callback(true);
+		}
+		callback(false);
+	},
+
+	getOidcLogoutTicketSecret: function() {
```
mikeTWC1984: What is the purpose of this function? It will always return secret_key, because this.server.config.get('secret_key') is guaranteed to exist if cronicle is up. 
You also generate that logout secret on setup, but it will never be used. Function comment doesn't make much sense to me.

lib/oidc.js
```diff
@@ -0,0 +1,798 @@
...
+	// Authenticate and decrypt an ID Token stored in the server-side session.
+	const packed = Buffer.from(String(value), 'base64url');
+	if (packed.length < 30 || packed[0] !== 1) throw new Error('Invalid encrypted ID Token');
+	const key = crypto.createHash('sha256').update(String(secret)).digest();
+	const decipher = crypto.createDecipheriv('aes-256-gcm', key, packed.subarray(1, 13));
+	decipher.setAuthTag(packed.subarray(13, 29));
+	return Buffer.concat([decipher.update(packed.subarray(29)), decipher.final()]).toString('utf8');
+}
+
+function buildSessionMetadata(oauth, idTokenClaims, userInfo, idToken, secret) {
```
mikeTWC1984: I don't see that function invoked anywhere

mikeTWC1984: And looks like this function is the only consumer of encryptSessionToken function and logoutSecret function I mentioned.

mikeTWC1984:
Also, going back to limiting usage of localhost. Does it serve any purpose rather than sanity check of using dev like config in prod? My point here is - running oidc/storage provider/etc on localhost is something I gonna use for dev occasionally. Unless needed I don't want to overconfigure cronicle with stuff I may not remember and figuring out how/what  to uncheck. 
So if using localhost causing any security risk I'm fine to keep it. If it just sanity - not sure about it. Or at least I need some way to disable all things like that in one shot (e.g. respect  --debug flag or some env variable)

frankii91:
You are right that two separate localhost flags overconfigure this use case. However, I would prefer not to enable HTTP automatically or connect it to --debug.
A localhost URL only identifies where a particular client connects. It does not prove that the provider is bound exclusively to a loopback interface. The meaning also differs by context: JWKS is fetched from the Cronicle host, while logout redirects are followed by the user's browser, and Docker has a separate loopback namespace.
I propose keeping one top-level oauth.allow_http_localhost setting and using it for issuer, JWKS, end-session and post-logout URLs. I will remove oauth.logout.allow_http_localhost. This provides one switch for the complete local development setup while keeping the non-standard HTTP exception explicit.
oauth.insecure will remain independent because it only controls TLS certificate verification.

mikeTWC1984:
OK, so let's have 1 localhost config and insecure separately as you said. I'm good to proceed with that proposal
But out of curiosity, I still do not understand issue with localhost to be honest. Can you give any practical example of what can go wrong? I understand if cronicle runs on hostA and client is on hostB this won't work, but is it any worse than just typo in configuration (I was able to broke OIDC just by missing / at the end of url)

mikeTWC1984:
OK, I'm pretty much done with review should be good to merge, once you reply to my review notes (on oidc.js)

frankii91:
Now I don't really know what's going on... what notes?

mikeTWC1984:
Don't you see it above?
image

I add comments on the files you change, but it should pop up on this discussion, I'd think you should get some notification about pending review or so

I left few comments like this on your other PRs

frankii91:
The comments in your screenshot are marked as Pending. GitHub keeps pending review comments visible only to the reviewer until the review is submitted. Could you please click “Review changes” and “Submit review”? Then I’ll be able to see and reply to all the inline comments.

frankii91:
Thanks, I addressed both review notes in c806139.

`secret_key` is required by a running Cronicle, so the generated process fallback had no useful runtime path. It was added defensively in the initial version and then carried into the mixin refactor; the comment added there was misleading. I removed the generated fallback and getter and now use `secret_key` directly.

`buildSessionMetadata` was invoked, but the alias made the call difficult to find. It was exported as `buildOidcSessionMetadata` and called from the OAuth callback in `user.js`; its result is merged into the session before `storeOidcSession`. It stores the provider identifiers and encrypts the ID Token when RP-Initiated Logout is enabled. I aligned the local and exported names and clarified the comment.

The full suite passes: 89/89 tests and 364 assertions.

mikeTWC1984:
OK, finally merging it.  Great work, thanks.

frankii91:
One final correction to my earlier one-flag proposal. After re-checking the exact call paths, I think the two existing localhost flags should remain separate. The merged code already keeps them separate, so no code change is needed.

One concrete configuration where this distinction matters is:

```json
{
  "issuer": "https://sso.example.com/application/o/cronicle/",
  "jwks_url": "http://127.0.0.1:9000/application/o/cronicle/jwks/",
  "allow_http_localhost": true,
  "logout": {
    "enabled": true,
    "end_session_url": "https://sso.example.com/application/o/cronicle/end-session/",
    "post_logout_redirect_uri": "https://cronicle.example/",
    "allow_http_localhost": false
  }
}
```

Here the provider keeps its public HTTPS issuer, while Cronicle deliberately fetches JWKS directly from a local listener instead of routing that server-to-server request through a reverse proxy or tunnel. The browser-facing logout URLs remain public HTTPS.

The top-level flag therefore permits loopback HTTP for server-side issuer/JWKS validation, while the nested flag permits loopback HTTP in the browser logout flow. Combining them would mean that enabling a valid local JWKS route also removes the separate validation guard for a browser logout URL. For the browser, `localhost` means the user's machine, and the end-session redirect can include `id_token_hint`.

The reverse configuration is also possible: issuer and JWKS remain on public HTTPS, while a Cronicle instance opened locally needs an HTTP loopback `post_logout_redirect_uri`. In that case only the nested logout flag is needed.

So this distinction is about where each URL is consumed, not whether the deployment is development or production. `oauth.insecure` remains separate because it only controls HTTPS certificate verification for token and UserInfo requests.

PR #251, opened 2026-08-15
fix(freebsd): restore job resource monitoring

frankii91:
## Summary

- keep the existing default `ps` command on Linux, macOS, and other platforms
- use the portable process query only on FreeBSD
- keep the existing output parser, Windows path, and `ps_monitor_cmd` override unchanged

## Problem

On FreeBSD, `/bin/ps -eo "ppid pid %cpu rss"` exits with an error. Cronicle then logs `Failed to exec ps` and does not update CPU or memory metrics for the server and active jobs.

## Fix

Keep `/bin/ps -eo "ppid pid %cpu rss"` as the default command. Only when `process.platform == 'freebsd'`, use `/bin/ps -axo ppid=,pid=,%cpu=,rss=`. The FreeBSD command selects all processes and emits the same four numeric fields without headers, so the existing parser remains unchanged.

## Validation

- verified the FreeBSD command on FreeBSD/amd64
- `npm ci`
- `node --check lib/job.js`
- `git diff --check`

The existing job lifecycle tests call `monitorServerResources` and verify that CPU and memory metrics are recorded.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -2128,11 +2128,11 @@ module.exports = Class.create({
 	monitorServerResources: function (callback) {
 		// monitor local CPU and memory for all active jobs (once per minute)
 		// shell exec to get running process cpu and memory usage
-		// this works on at least: OS X, Fedora, Ubuntu and CentOS
+		// this works on at least: FreeBSD, OS X, Fedora, Ubuntu and CentOS
 		var self = this;
 		var now = Tools.timeNow();
 
-		var defCmd = '/bin/ps -eo "ppid pid %cpu rss"' 
+		var defCmd = '/bin/ps -axo ppid=,pid=,%cpu=,rss='
 		var kb = 1024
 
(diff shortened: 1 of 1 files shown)


frankii91:
One thing I'm unsure about: the Docker / build (pull_request) check failed again. I may be missing something in how this check is expected to run. The project test suite passes locally: 55/55 tests, 364 assertions.

frankii91:
I checked the failed Docker workflow. GHCR builds successfully; only Docker Hub fails because that step runs on `pull_request`, where fork PRs do not receive `DHUB_USER` / `DHUB_TOKEN`.

Could you consider adding `if: github.event_name != 'pull_request'` to the Docker Hub step? PR images would still be validated by GHCR, while Docker Hub publishing would remain limited to push/release.

mikeTWC1984:
do it only for bsd, don't change default command

PR #252, opened 2026-08-15
fix(freebsd): fail safely when getnode is unsupported

frankii91:
## Summary

- stop `getnode.sh` before changing an existing installation when run on FreeBSD
- explain that Node.js must be installed through the FreeBSD package manager
- locate Bash through `PATH`, where FreeBSD packages install it

## Problem

Node.js does not publish FreeBSD binaries in its official release archive. On FreeBSD/amd64, `get_arch` returns no platform, but the script continues with a URL ending in `node-v20.16.0-.tar.xz`.

Before that download fails, the script removes the existing `nodejs` directory. It then continues after the failed download and can print `Node.js installed` with exit code 0.

## Fix

Detect FreeBSD while resolving the archive platform and return a clear error. Propagate that result before constructing the URL or removing the existing installation. Use `/usr/bin/env bash` because packaged Bash lives outside `/bin` on standard FreeBSD. Linux and macOS archive selection remains unchanged.

## Validation

- confirmed the target host reports `FreeBSD/amd64`
- reproduced the old behavior: invalid URL, existing-install sentinel deleted, exit code 0
- verified the fixed behavior: clear error, exit code 2, existing-install sentinel preserved
- verified Linux/x64 still resolves to `node-v20.16.0-linux-x64.tar.xz`
- verified `/usr/bin/env bash` on the FreeBSD target
- `bash -n bin/getnode.sh`
- `git diff --check`

### DIFF
--- a/bin/getnode.sh
+++ b/bin/getnode.sh
@@ -7,10 +7,15 @@ DEFAULT_VERSION="20.16.0"
 NODEJS_VERSION="${1:-$DEFAULT_VERSION}"
 
 get_arch () {
-	local arch
+	local arch os os_name
 	arch=$(uname -m)
-    os="linux"
-    if [[ "$(uname)" == "Darwin" ]]; then os="darwin"; fi 
+	os_name=$(uname -s)
+	if [[ "$os_name" == "FreeBSD" ]]; then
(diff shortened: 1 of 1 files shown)


mikeTWC1984:
LGTM

PR #253, opened 2026-08-15
fix(freebsd): use portable bash shebangs

frankii91:
## Summary

Use `#!/usr/bin/env bash` for the two Bash entrypoints that still hardcode `/bin/bash`.

## Problem

FreeBSD does not include Bash in the base system. The package installs it under `/usr/local/bin/bash`, so `bundle` and `bin/manager` can fail before execution even when Bash is installed.

Both scripts use Bash-specific syntax, so changing them to `/bin/sh` would not be correct. Resolving Bash through `PATH` matches the existing script examples in this repository and works on FreeBSD, Linux, and macOS.

`bin/getnode.sh` is handled separately in #252 to avoid overlapping changes.

## Validation

- `bash -n bundle bin/manager`
- both files passed `bash -n` through `/usr/bin/env bash` on FreeBSD
- both executable help paths resolve the new shebang locally
- `git diff --check`

No JavaScript or runtime behavior is changed.

### DIFF
--- a/bin/manager
+++ b/bin/manager
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 HOMEDIR="$(dirname "$(cd -- "$(dirname "$(readlink -f "$0")")" && (pwd -P 2>/dev/null || pwd))")"
 
 usage() {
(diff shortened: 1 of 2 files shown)


mikeTWC1984:
Ok, time to review all your other PRs. Agree on this one

PR #256, opened 2026-08-15
Prevent disabled LDAP group users from reactivating

frankii91:
## Summary

- reject locally disabled LDAP group users before refreshing their account
- keep the local `active` control authoritative while leaving password and lockout policy to LDAP/AD
- add regressions for disabled group users and stale local password-reset flags

## Why

Group-auth login refreshed and replaced the stored user before checking `active`. A disabled user with valid LDAP credentials could therefore be written back with `active: 1` and log in again.

The early check applies only to existing group-auth users while LDAP group authentication is enabled. A stale `force_password_reset` value does not block a group user or show a Cronicle password-reset message; LDAP/AD remains responsible for authenticating and locking directory accounts.

New LDAP users and local-password users keep their existing flow.

## Tests

- before the original fix: `55/56` tests passed; the regression showed that LDAP was called and the stored account became active
- current branch: `57/57` tests passed (`374` assertions)
- `node --check lib/user.js`
- `node --check lib/test.js`
- `git diff --check`

### DIFF
--- a/lib/user.js
+++ b/lib/user.js
@@ -815,6 +815,14 @@ module.exports = Class.create({
 
 		let useGroupAuth = (user || {}).group_auth
 
+		// Keep local account controls authoritative over LDAP refreshes.
+		if (allowGroupAuth && useGroupAuth && user.force_password_reset) {
+			return self.doError('login', "Account is locked out.  Please reset your password to unlock it.", callback);
+		}
+		if (allowGroupAuth && useGroupAuth && !user.active) {
+			return self.doError('login', "User account is disabled: " + params.username, callback);
+		}
+
 		if (allowGroupAuth) {
 			if (!user || useGroupAuth) {
 				// create or verify LDAP group users
--- a/lib/user.js
+++ b/lib/user.js
@@ -815,10 +815,7 @@ module.exports = Class.create({
 
 		let useGroupAuth = (user || {}).group_auth
 
-		// Keep local account controls authoritative over LDAP refreshes.
-		if (allowGroupAuth && useGroupAuth && user.force_password_reset) {
-			return self.doError('login', "Account is locked out.  Please reset your password to unlock it.", callback);
-		}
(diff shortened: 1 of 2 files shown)


lib/user.js
```diff
@@ -815,6 +815,14 @@ module.exports = Class.create({
 
 		let useGroupAuth = (user || {}).group_auth
 
+		// Keep local account controls authoritative over LDAP refreshes.
+		if (allowGroupAuth && useGroupAuth && user.force_password_reset) {
+			return self.doError('login', "Account is locked out.  Please reset your password to unlock it.", callback);
+		}
+		if (allowGroupAuth && useGroupAuth && !user.active) {
```
mikeTWC1984, line 819: allowGroupAuth && useGroupAuth  should allow user creation. When you add !user.active I think you'll break this. Should also check if user exist

mikeTWC1984: And do we really need to check for allowGroupAuth && useGroupAuth ? shouldn't be just (user && !user.active)

mikeTWC1984:
Also, just realized - user.force_password_reset is only related to local user. Technically it should never be on the AD user object (type of user cannot be changed from local to AD or wise versa ). Even if by some reason AD user would get this flag - I'd rather ignore it and let AD handle it (I guess it lock for N minutes after 3 retries). AD user should not see "reset password" message from cronicle

lib/user.js
frankii91: User creation is not blocked because useGroupAuth is falsy when no local user exists. However, I agree that making the existence check explicit would be clearer, and I’ll add a regression test for first-time LDAP group-user creation.
I would keep the group-auth guards rather than changing this to user && !user.active. The broader condition would change the existing groups.allow = false path from the deliberately vague login error to an explicit “account is disabled” response, revealing that the account exists. The early check is only needed for existing group-auth users whose LDAP refresh would otherwise overwrite the local active: 0 state.

mikeTWC1984: OK, agree

PR #257, opened 2026-08-15
fix: restrict unprivileged run_event overrides

frankii91:
## Problem

`app/run_event` merged caller-controlled query/body fields into the stored event.  A principal allowed to run, but not edit, an event could therefore override execution fields outside the documented ad-hoc inputs.  The same merge also accepted prototype-control keys from raw JSON.

## Change

- Restrict run-only callers to `now`, `arg`, `args`, and opaque `post_data`.
- Preserve the existing nested override behavior for users who can edit events.
- Treat `chain_data` and `memo` as scheduler-owned state for every role, incorporating the maintainer feedback on this PR.
- Strip caller-supplied authentication and provenance fields, then derive provenance on the server.
- Merge own properties only and reject `__proto__`, `prototype`, and `constructor` at the request boundary.

## Compatibility

The normal UI Run Now flow, workflow child arguments, GitHub/event-token `post_data`, and API-key authentication remain supported.  Stored event configuration remains the source of execution policy for run-only principals.

## Validation

- Real HTTP regressions cover event-token and run-only override attempts, editor compatibility, raw JSON prototype payloads, and server-derived provenance.
- Repeat regressions cover positive and negative caller values plus `enabled=false`; a real launch-capacity failure verifies the durable `global/event_queue/<id>` record retains the stored event policy instead of caller input.
- Full `npm test`: 93/93 tests, 432 assertions.
- `node --check` and `git diff --check`: pass.

The issues and initial patches were proposed by a Codex Ultra security review, including the repeat-loop report in [frankii91/cronicle-edge#16](https://github.com/frankii91/cronicle-edge/pull/16). This version was manually re-analyzed, covers the complete run-only override boundary instead of adding a field-specific guard, adds real launch and durable-queue regression coverage, and is submitted for maintainer review in line with our prior discussion.

### DIFF
--- a/lib/api/event.js
+++ b/lib/api/event.js
@@ -415,15 +415,29 @@ module.exports = Class.create({
 
 	api_run_event: function (args, callback) {
 		// run event manually (via "Run" button in UI or by API Key)
-		// can include any event overrides in params (such as 'now')
+		// can include authorized runtime overrides such as now, arg and post_data
 		var self = this;
 		if (!this.requiremanager(args, callback)) return;
 
-		// default behavor: merge post params and query together
-		// alt behavior (post_data): store post params into post_data
-		var params = Tools.copyHash(args.query, true);
-		if (args.query.post_data) params.post_data = args.params;
-		else Tools.mergeHashInto(params, args.params);
+		// Default behavior: merge request params and query using own properties only.
+		// Never assign prototype-control keys into a normal object.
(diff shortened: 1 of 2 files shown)


mikeTWC1984:
OK, good point.. I think we need to remove params.memo for any role, since it meant to be generated at job runtime. 
I'm 99% sure same applies to chain_data, but let's keep it the way you proposed. Don't want to break some internals accidentally.

PR #259, opened 2026-08-15
fix(freebsd): use portable ps resource query

frankii91:
## Summary

- keep the existing `/bin/ps -eo "ppid pid %cpu rss"` query on Linux, macOS, and other established platforms
- use the FreeBSD-specific query only when `process.platform == 'freebsd'`
- request each FreeBSD numeric column with a separate `-o` option and normalize its locale with `LC_ALL=C`
- keep the existing process parser unchanged

## Why

On FreeBSD 14.3, `-e` adds environment information instead of selecting all processes, so the original command does not provide the complete process tree needed for CPU and memory monitoring. A localized FreeBSD `%CPU` value can also use a decimal comma, while the existing parser expects a decimal point.

`LC_ALL=C /bin/ps -ax -o ppid= -o pid= -o %cpu= -o rss=` returns the four headerless numeric columns expected by the parser. The platform guard leaves the longstanding command and behavior untouched everywhere else.

## Tests

- Production FreeBSD S21 with a Polish numeric locale: the FreeBSD command produced `29` parsed rows and found the current Node process with CPU and memory values.
- Linux: the unchanged default command produced `978` rows accepted by the existing parser.
- Full suite: `55/55` tests passed (`364` assertions).
- `node --check lib/job.js`
- `git diff --check`

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -2128,11 +2128,11 @@ module.exports = Class.create({
 	monitorServerResources: function (callback) {
 		// monitor local CPU and memory for all active jobs (once per minute)
 		// shell exec to get running process cpu and memory usage
-		// this works on at least: OS X, Fedora, Ubuntu and CentOS
+		// this works on at least: FreeBSD, OS X, Fedora, Ubuntu and CentOS
 		var self = this;
 		var now = Tools.timeNow();
 
-		var defCmd = '/bin/ps -eo "ppid pid %cpu rss"' 
+		var defCmd = '/bin/ps -axo ppid=,pid=,%cpu=,rss='
 		var kb = 1024
 
 		if(process.platform == 'win32') {
--- a/lib/job.js
+++ b/lib/job.js
@@ -2132,7 +2132,7 @@ module.exports = Class.create({
 		var self = this;
 		var now = Tools.timeNow();
 
-		var defCmd = '/bin/ps -axo ppid=,pid=,%cpu=,rss='
+		var defCmd = '/bin/ps -ax -o ppid= -o pid= -o %cpu= -o rss='
 		var kb = 1024
 
 		if(process.platform == 'win32') {
--- a/lib/job.js
+++ b/lib/job.js
@@ -2128,12 +2128,17 @@ module.exports = Class.create({
 	monitorServerResources: function (callback) {
 		// monitor local CPU and memory for all active jobs (once per minute)
 		// shell exec to get running process cpu and memory usage
-		// this works on at least: FreeBSD, OS X, Fedora, Ubuntu and CentOS
+		// this works on at least: OS X, Fedora, Ubuntu and CentOS
 		var self = this;
 		var now = Tools.timeNow();
 
-		var defCmd = 'LC_ALL=C /bin/ps -ax -o ppid= -o pid= -o %cpu= -o rss='
+		var defCmd = '/bin/ps -eo "ppid pid %cpu rss"'
 		var kb = 1024
+
+		if(process.platform == 'freebsd') {
+			// FreeBSD -e prints the environment instead of selecting all processes.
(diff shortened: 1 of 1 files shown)


mikeTWC1984:
What problem does it solve?

frankii91:
Thanks for asking. Your question made me re-check the exact runtime behavior, and it exposed a mistake in my initial testing method.

I originally tested through a non-interactive SSH command with `LC_ALL=C`, while my normal interactive FreeBSD session has a TTY and localized numeric output. This produced different results.

The original command does run, but on FreeBSD 14.3 `-e` adds environment information instead of selecting all processes. It therefore does not provide the complete process tree Cronicle needs for CPU and memory monitoring. Additionally, `%CPU` may be returned as `0,0`, while the existing parser expects `0.0`.

I updated the command to:

`LC_ALL=C /bin/ps -ax -o ppid= -o pid= -o %cpu= -o rss=`

I verified it on the actual FreeBSD host with a Polish numeric locale.

mikeTWC1984:
OK, but can we do it only if platform is freebsd though, not for all. That is small but important part, since it worked for years I don't want to touch it unless it's broken

frankii91:
Sure, that makes sense. I restored the existing `ps` command as the default and now select the new command only when `process.platform == 'freebsd'`. Linux, macOS, and the other existing platforms are unchanged.

The full suite still passes (`55/55`, 364 assertions), as do the syntax and diff checks. I also updated the PR description to reflect the platform-specific scope.

mikeTWC1984:
OK, good. One last thing - I guess setting LC variable needed to make it work over ssh, right? I don't think it's needed when cornicle engine will be invoking it.

frankii91:
Good question. It was not required by SSH itself. SSH only exposed the issue because my interactive session used a Polish locale.
I checked the currently running Cronicle process on the FreeBSD host. It has no LANG or LC_* variables, so in that exact environment ps already uses the C locale and the prefix is currently redundant.
However, Cronicle calls cp.exec() without an explicit env, so the command inherits the Cronicle process environment. If Cronicle is started from a localized shell or service configuration, FreeBSD ps inherits that locale. On the same host I verified:
LC_ALL=pl_PL.ISO8859-2  -> 13,4
LC_ALL=C                -> 13.4
The existing parser accepts the decimal point but not the comma. I therefore kept LC_ALL=C in the FreeBSD-only default command to make its output deterministic regardless of how Cronicle was started. It does not affect Linux, macOS, Windows, or a custom ps_monitor_cmd.

mikeTWC1984:
Can't you set that variable before starting up cronicle though? Also - in config (either config.json or UI) you can set property like env.LC_ALL, it will inject this  variable at  job runtime. 
 But anyway, if you say it's needed for freeBSD I have no issue adding it.

frankii91:
Tak, ustawienie LC_ALL=C przed uruchomieniem demona Cronicle również by zadziałało. Jednak env.LC_ALL ustawione w konfiguracji lub UI jest przekazywane do uruchamianych zadań, natomiast monitorServerResources() uruchamia ps bezpośrednio z silnika Cronicle przez cp.exec() i nie korzysta ze środowiska zadania. Pozostawienie tego prefiksu w domyślnym poleceniu wyłącznie dla FreeBSD zapewnia przewidywalne działanie monitorowania bez konieczności dodatkowego konfigurowania każdego wdrożenia. Dziękuję za potwierdzenie - pozostawię rozwiązanie w obecnej formie.

frankii91:
Yes, setting LC_ALL=C before starting the Cronicle daemon would work as well. However, env.LC_ALL from config/UI is injected into launched jobs, while monitorServerResources() runs ps directly from the Cronicle engine via cp.exec() and does not use the job environment. Keeping the prefix in the FreeBSD-only default command makes this engine-level monitoring deterministic without requiring additional deployment configuration. Thanks for confirming - I’ll keep it as is.

PR #260, opened 2026-08-15
fix(freebsd): locate Bash through PATH

frankii91:
## Summary

- locate Bash through `PATH` in `getnode.sh`, `manager`, and `bundle`
- leave all script logic unchanged

## Why

FreeBSD does not include Bash in the base system, and the package installs it outside `/bin`. The fixed `/bin/bash` shebang therefore fails on a standard FreeBSD installation even when Bash is installed and available in `PATH`.

FreeBSD documents the packaged shell as `/usr/local/bin/bash`: https://docs.freebsd.org/en/books/handbook/basics/#shells

## Tests

- `bash -n bin/getnode.sh bin/manager bundle`
- `/usr/bin/env bash -c 'exit 0'`
- `git diff --check`


mikeTWC1984:
OK, make sense

Issue #13, opened 2022-02-15
Error: Failed to locate Secret: globalenv


dheise-rheinenergie-trading-com:
I can't open the tab "admin" - "configs". I only get the error message "Error: Failed to locate Secret: globalenv". Any idea what is missing?

mikeTWC1984:
I guess you are running it on top of existing data folder? 
You need to init global/secrets/globalenv object in storage to make it work ( it supposed to be created during initial set up which you don't run if you have data folder already).  It might be possible create it with storage-cli, I'll check. Maybe I can add an option to auto generate this objrct

In general I meant this fork to run as a fresh install, and then  import jobs from original cronicle via import api to avoid any incompatibility issues.

dheise-rheinenergie-trading-com:
Hi Mike,

thanks for the hint with the storage-cli.js. I was able to transport the settings from a scratch-installation to my system.

1. create the list: ./storage-cli.js list_create global/secrets
2. edit the item 0: ./storage-cli.js edit global/secrets/0 and paste:
```
{
        "type": "list_page",
        "items": [
                {
                        "id": "globalenv",
                        "encrypted": false,
                        "target": null,
                        "form": "props",
                        "data": "# dotenv style key/value pairs representing env variables\nmyvar = some_value"
                }
        ]
}
```
3. edit the list and set the length from 0 to 1: ./storage-cli.js edit global/secrets
4. error is gone!

PR #264, opened 2026-08-17
Handle missing global/secrets in migrated Cronicle storage

fmaass:
After migrating storage from classic Cronicle 0.9.x, the server crash-looped with `Emergency Shutdown` during login/session resume and when a scheduled job launched.

Classic 0.9.x storage can lack the `global/secrets` list and its `globalenv` item. Fresh Edge setups seed both from `sample_conf/setup.json`, but `bin/storage-cli.js setup` exits when `global/users` already exists ("Storage has already been set up"), so an existing dataset is not backfilled. A migrated 0.9.x dataset can therefore lack both records.

Two unguarded dereferences, each with an already-guarded sibling a few lines away:

**1. `lib/engine.js`, `beforeUserLogin`** — `listGet('global/secrets')` yields `null` when the list is absent, and the `secrets` branch calls `.forEach` on the copy directly. The `plugins`/`schedule` branch two lines above already wraps the identical call in `if (Array.isArray(itemCopy))`.

```
Emergency Shutdown: TypeError: Cannot read properties of null (reading 'forEach')
```

**2. `lib/job.js`, job launch** — `globalenv = item;` takes the `listFind` result unfiltered, and `globalenv.data` is read further down the same function. The next two fetches in that block, `cat_secret` and `plug_secret`, both already use `item || {}`.

```
Emergency Shutdown: TypeError: Cannot read properties of null (reading 'data')
```

This change mirrors the existing guards nearby: it only iterates the secrets result when it is an array and defaults a missing `globalenv` item to `{}`, matching how category and plugin secrets already degrade.

My deployment runs the bundled build, so the raw traces point into `bin/cronicle.js` rather than `lib/*.js`; the sites above are the corresponding source lines. The bundle is regenerated from `lib/` by the Docker build, so the source guards reach the shipped artifact.

Verification, done against a copy of the migrated dataset rather than the live instance:

- **before** — reproduced both crashes: the first on a valid session resume, the second at the next scheduler minute-tick that launched a job.
- **after** — creating the `global/secrets` list and adding a `globalenv` item, exactly as `setup.json` seeds them, stops both crashes. That is what pins the trigger to the missing records rather than to anything else in the migrated data.
- For the patch, I ran `node --check lib/engine.js && node --check lib/job.js` and `npm test` (55/55 pass on Node 22). I did not rebuild the bundle or run the patched code against the migrated dataset. The test suite seeds storage from `sample_conf/setup.json`, so it does not cover the missing-record case; I did not add a unit test.

Issue #13 concerns a separate admin-UI error and is unrelated to these crashes.

### DIFF
--- a/lib/engine.js
+++ b/lib/engine.js
@@ -584,7 +584,7 @@ module.exports = Class.create({
 						if (Array.isArray(itemCopy)) itemCopy.forEach(obj => delete obj.secret)
 					}
 					if (name == 'secrets') {
-						itemCopy.forEach(obj => delete obj.data)
+						if (Array.isArray(itemCopy)) itemCopy.forEach(obj => delete obj.data)
 					}	
 					args.resp[name] = itemCopy || [];
 					callback();
--- a/lib/job.js
+++ b/lib/job.js
@@ -118,7 +118,7 @@ module.exports = Class.create({
 			},
 			function (callback) {
 				self.storage.listFind('global/secrets', { id: 'globalenv' }, function (err, item) {
-					globalenv = item;
+					globalenv = item || {};
 					callback(err);
 				});
 			},


mikeTWC1984:
@fmaass did you migrate just by running cronicle-edge on top of existing cronicle data folder, or using "import" button on Schedule page (from old cronicle backup output)?

fmaass:
@mikeTWC1984 oh, oops. I ran cronicle-edge on top of the existing 0.9.99 data folder — no Import button, no backup-file re-import.
I was annoyed with the missing sso/oidc function and found this fork - wasn't aware of the import function :D

mikeTWC1984:
OK. Were you able to run it on top of old /data folder with changes you made though? If so, try to run it for a while and see if it works in the long run. You probably see some other quicks or UI issues here and there.
I haven't touch migration part in a while, as far as I remember it won't work as a drop in replacement, but perhaps it can be achieved with few tweaks like you suggest.

fmaass:
Yes — I’ve now run it both ways and can be more precise than in my earlier description.

**Patched build on the raw old /data (test copy).** I built this branch’s image and ran it against a byte-identical copy of the pre-migration 0.9.99 data folder, with nothing seeded. The unpatched image remains healthy while idle but dies on the first login (`Emergency Shutdown: TypeError: Cannot read properties of null (reading 'forEach')` — re-verified on the same copy). By contrast, the patched build returns `code 0` with `secrets: []` for both `login` and `resume_session` and stays up. I then left it running with a once-a-minute `resume_session` probe: 50/50 successes over ~50 minutes, no restarts and no crash lines in the log. I also verified that both guards made it into the generated `bin/cronicle.js` bundle, not just `lib/`. The note in the PR description ("did not run the patched code against the migrated dataset") is therefore now closed.

**One correction on the job-side crash.** On fully raw migrated data, the `job.js` guard is never reached: every job launch fails earlier and cleanly with `Failed to launch event: Failed to fetch key: global/secrets: File not found`, identically with and without the patch, because the callback propagates the storage error before the deref. The `reading 'data'` crash fires once `global/secrets` exists *without* a `globalenv` item — the intermediate state reached if the list is backfilled before the item. So, to be exact: the two guards stop the crashes, but a migrated instance additionally needs `global/secrets` + its `globalenv` item backfilled (as `sample_conf/setup.json` seeds them) before jobs will launch. That suggests the cleanest "few tweaks" route to a real drop-in: have `storage-cli setup` (or startup) backfill missing `global/*` records on existing storage instead of exiting with "Storage has already been set up". I’m happy to attempt that in a follow-up PR if that direction is welcome.

**Long run on the live instance.** My production instance has been running on the in-place migrated data folder since yesterday, with the two records seeded manually via `storage-cli` and the upstream image unpatched: one continuous process, zero restarts, and scheduled jobs firing normally every 10 minutes. The migrated data came through well — job history back to March renders fully, old job logs load with UTF-8 intact, and events, categories, plugins and API keys are all intact.

**Quirks noticed so far:**

- Every Shell Script event’s edit view throws `Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')` in `setScriptEditor` (`combo.min.js`), which aborts init and leaves the script-source editor unthemed — white CodeMirror on a dark-mode page. I can’t yet tell whether this is migration-related or general; I can file it separately with details if useful.
- Job-log line prefixes changed format across the migration (`[2026/08/16 03:45:00]` local time → ISO UTC), so old and new logs of the same event use two formats. Cosmetic.
- Nothing else yet — I’ll report back if longer runtime shakes anything out.

fmaass:
Following up on the discussion above, I opened three self-contained PRs:

- **#265** adds the `storage-cli.js backfill` command — the drop-in migration you suggested might work with a few tweaks. It creates only the `global/*` records missing from a migrated folder, exactly as `setup` would seed them, and never touches anything that exists. The manual storage surgery I described earlier is therefore no longer needed: backfill the folder once, and jobs launch.
- **#266** fixes the event-editor quirk from my last comment and corrects my diagnosis there, which was wrong on two counts. It is not migration damage: a fresh install reproduces the identical trace with any plugin that has a `script` param but no `lang`/`theme` selects. That is exactly the shape of a shellplug record carried over from classic Cronicle, hence it affecting every migrated event. It is also two independent defects with the same trigger, not one: the unguarded `addEventListener` calls on the absent selects, which also left the UI stuck on the previous event when navigating, and an unset `theme` param separately skipping the dark-mode mapping — that alone caused the white editor. Both are fixed.
- **#267** addresses the other quirk from that comment: one event's job logs switching from `[yyyy/mm/dd hh:mi:ss]` local time to ISO UTC at the migration date. It adds an opt-in `stamp_format` select on the shellplug offering the classic local stamp, while leaving the current ISO output as the default (I saw that the ISO change was deliberate in v1.7.1, so this is strictly additive).

Status update on the long run: still one continuous process, no restarts, jobs firing normally.

fmaass:

PR #265, opened 2026-08-18
Compatibility fixes for installs migrated from classic Cronicle

fmaass:
Compatibility fixes for a data folder migrated from classic Cronicle, reworked along the lines discussed below. The branch is rebased on current main and has one commit per item.

- **Missing global lists.** When a server becomes manager, it checks an explicit list of the lists this version requires (`global/secrets`, `global/conf_keys`) and creates only those whose key is absent. It seeds the non-optional setup entries (`globalenv` for secrets, nothing for conf_keys), matching a minimal setup. Existing records are never touched. An empty required list is a valid end state, so there is no rollback logic. A failed seed is logged and retried on the next manager start. It runs manager-side because cluster nodes share storage.
- **One-time legacy plugin import.** A folder whose `global/server_groups` records carry the classic `master` flag and no `manager` field gets the edge-only stock plugins imported by id (`workflow`, `sshplug`, `sshxplug`, `dockerplug`, `kuberun`, `kubeyaml`, `kubelist`, `stdin`, `terminal`). It also gets copies of the shell and test plugins under new ids (`shellplug_v2`, `testplug_v2`, titles with " V2"). A candidate whose id or title already exists is skipped, and no existing plugin record is modified, so switching back to classic keeps working. Only after the plugin list has been read back and shows every import is `manager` copied from `master` onto each of those groups. That write is the fingerprint that stops the import from running again and makes eligibility permanent through the canonical field. Logged-in browsers are refreshed through the existing client-data broadcast.
- **Classic `master` groups are manager-eligible.** A small read-path helper falls back to `master` only when `manager` is absent, so a migrated node can become manager on its first boot and the two steps above can run. Once `manager` is present, it decides alone. Setting it to 0 on a stamped group therefore revokes eligibility even though the record still carries `master`.
- **Syntax and theme selectors derived from the script param.** The option lists are defined by the build. The event editor now synthesizes both selects whenever a plugin declares a `script` textarea param and ignores any lang/theme param a stored plugin still declares. One helper supplies the effective param list to the render path, the plugin-change defaults and `get_event_form_json`, so the selects that render are the ones that save. A stored value not on the list falls back to the default. A legacy event carrying only `default_lang` keeps its syntax, and plugins without a script param are unchanged. The stock plugins in `setup.json` drop their lang/theme params; `terminal` keeps its JavaScript default through a hidden `default_lang` param.
- **Null guards for migrated storage** and the **opt-in classic log timestamp** (`classic_stamp`, default off) are unchanged from the previous revision.

`npm test`: 107 of 107, 524 assertions. I also checked the editor change in a real browser against a local instance: a stored `python` value renders as python, an off-list value renders as shell, a plugin declaring only a `script` textarea renders both selects and saves `lang` and `theme` on the event, a plugin without a script param renders no selects, and a new terminal event opens in JavaScript.


mikeTWC1984:
OK, I gotta play around with this a bit. But looks good at the glance.
 But effectively, if you run it now, it will just create secret and config_key collection, right?

Technically we can try to check/add those collection on startup, making this seamless for users. Actually there could be something like this already, since I haven't had secrets and configs from day 1

mikeTWC1984:
Nah, looks like there is no logic to auto create secret list. I guess re-export was required to migrate from older versions, somehow cannot find release notes for this :)) 
 Anyway, if it's just a matter of adding missing collection, I'd try to add this logic in startup section of lib/egine.js? Can you do quick experiment?

fmaass:
Tried it and it works, so I changed the PR to do exactly that — the backfill command is dropped.

`createMissingRecords()` in lib/engine.js walks the same setup.json record list and creates only the `global/*` keys whose `storage.get` comes back NoSuchKey, seeding each one the way setup does. An existing key is skipped entirely, and any other read error leaves that record alone. `global/users` is excluded so a migrated install never gets a default admin account back.

I hung it off `gomanager()` rather than raw startup: in a cluster all nodes share storage, so the manager is the only node that may seed, and that's where engine.js already keeps its manager-only storage work.

I tested it by seeding a data folder with setup, deleting `global/secrets`, `global/conf_keys` and `global/categories`, then starting the engine — one "Created missing record:" line per key, contents match what setup writes, engine stays up. A restart creates nothing and every global record is byte-identical; a run against an intact folder is 8 reads and zero writes.

I also added two guards from review feedback. A per-process latch, because a failover chain that stalls past `manager_ping_timeout` can fire `gomanager` twice and both passes would see the key missing. And an undo for partial failures: `listCreate` writes the header before its first page and `listDelete` won't remove a header once a page is missing, so after a failed seed the key is checked under the same lock the list ops use, and the leftover header is removed only when it's provably the failed pass's own residue — a list header with no first page. Anything a concurrent writer put there survives, and the next manager start retries cleanly. npm test 60/60, including the missing-then-present path, overlapping calls, both partial-failure shapes, and a concurrent-writer scenario.

Two points where I'd like your preference:

- Records appear when a node becomes manager, which on a single node is roughly a minute after boot via failover, not at t=0. Jobs can't launch before a manager exists, so the failure I hit is covered, but the Secrets page would 404 for that first minute. Move it to unconditional startup instead?
- The engine reads the bundled sample_conf/setup.json (esbuild inlines it), while storage-cli setup reads conf/setup.json at runtime. If you'd rather have one source of truth, switching to an fs read is small.

My read is that with this in, the null guards in #264 become mostly belt-and-suspenders for the migration case — I'll leave #264 to you.

mikeTWC1984:
OK cool. I'll work on testing it. Can we incorporate all your other PRs in this PR, since they all about "compatibility", I want to test them all together.

mikeTWC1984:
OK, good.

On problem I just realized - in old cronicle "manager" is called master. So if just run edge on top of upstream cronicle - main node will fail to become a manager because it's checking for group.manager which is group.master. So fixing it on "check manager eligibility" let it start normally and it looks good at a glance.

So obviously, if yo run it this way - no new plugin will be available and shell plugin will miss some new stuff. See if you can also export workflow and patch shell plugins. At least. Perhaps we can patch/import all the plugins, but I'd say those two are important to get benefit of switching to edge.

mikeTWC1984:
Yeah, and I also still testing. I guess that manager/master issue may require some tweaks in other places. but good job so far

fmaass:
Thanks for digging into it this far.

I folded #266 and #267 in here as you asked, and #264 too since it's the same compatibility theme — each is preserved as its own commit, so this branch now has the missing-record seeding, #264's null-guards, the event-editor selector fix, and the hidden classic_stamp option.

On the master/manager finding: good catch, and it actually explains why the seeding looked quiet. Since createMissingRecords runs from gomanager(), a migrated node that never becomes manager never seeds. I added a small isManagerGroup() helper that treats a group as manager-eligible when either `manager` or `master` is set, applied at the eligibility scan and the election candidate loop. It's read-only — I don't rewrite the stored server_groups record (this branch's whole approach is to never touch an existing record), so the write path stays on the canonical `manager` and a group re-saved through the UI normalizes itself. That's pushed here now; npm test 61/61.

On patching the plugins: agreed that's the real payoff of moving to edge, and I'm happy to build it — but it's a different kind of operation from everything above. The migrated folder already has a global/plugins list, and so far this branch only ever creates records that are missing, never edits one that exists. Patching plugins means editing that existing list, so before I build it I'd like your call on how far it should go:

- add-only: create any plugin that's missing (e.g. workflow) and add missing params to existing ones (e.g. the shell plugin's newer fields), but never overwrite a value someone has customized; or
- full sync: replace the stock plugins with edge's current definitions — simpler, but it clobbers any local edits to those plugins.

My instinct is add-only for the two you named (shell + workflow), so nobody loses customizations. Which way do you want it?

mikeTWC1984:
Some new thoughts.

1) Language and theme lists are pre-defined by specific cronicle build. So technically it's pointless to have it as independent parameter. Thinking to turn it into "pseudo" parameter. Say if plugin has parameter script of type textbox - we should inject pre-defined list on lang/theme (regardless if plugin already have such params or not).  I'd even remove lang/theme from shellplug-like plugins from setup.json. On UI side - we also need to check current value of lang/theme stored on event, if it's not on the predefined list should fallback to default.

I think your change is very close to it already, just have some pretty complex if/else analysis. I think checking just for script parameter might kill all birds with one stone

2) The "createMissingPart" kinda repeat initial setup step?  I think it's a bit to broad for the task.
Maybe it make sense, but I'd feel more safe if we just do "integrity check" first - e.g. define list of required lists (secrets,configs) ,then check if those exist and if not - create. This would be useful for ether switching from legacy to edge, and if we do some major edge upgrade (adding some new list)
 The other part could be only for legacy compatibility, which would be porting new plugins. I'd also do it once (need to leave some fingerprints on completion)

 After giving another thought - I think we should not modify any existing plugin on legacy setup. We can import workflow/sshx/docker/kube plugins, but for test and shell - create a copy with V2 in the name. This way user can switch back to old cronicle with no issues.
  I think we can detect "initial legacy switch" if we see main server has master=1 and manager=undefined. Then we can set manager=1 to prevent this process on the next start.
I wouldn't touch other stuff from setup.json for now

Anyway, just thinking out loud for now. There could be some flaws in my proposals too. Give it a though

fmaass:
Thanks, this all makes sense to me. Here is how I'd take each point, plus one hole I found in my own change while checking it against your proposal.

**1. lang/theme as pseudo params.** Agreed, and it's simpler than what I have. Neither value is read anywhere server side (I grepped lib/ and bin/), so they are purely an editor concern. I'd add one helper in Schedule.class.js to derive the effective param list for a plugin: drop any declared `lang`/`theme` param, and if the plugin has a `script` param of type textarea, append the two selects from a single build-defined list. A declared `default_lang` (or a hidden `lang` value) still seeds the default. The same helper feeds the render, the defaults applied on plugin change, and `get_event_form_json`; a stored value not on the list falls back to the default. That last part also fixes the hole: my current commit only patched the render side, while `get_event_form_json` still walks `plugin.params`, so for a plugin that declares no lang/theme the injected selects render but never save. One shared helper fixes that. Removing lang/theme from the setup.json plugins is then safe; existing installs keep them in the stored record and the helper ignores them. The stock lists differ per plugin today (shellplug has toml and light, the ssh/docker ones have json and props), so the shared list would be the union.

**2. Integrity check instead of replaying setup.** Agreed. I'll narrow it to an explicit array of required lists. `global/secrets` and `global/conf_keys` are the only global records edge has that a classic folder doesn't, so that is the whole list for now; a future edge version that adds a list adds one entry. For each, if the key is missing, create the list and push only the non-optional seed entries: `globalenv` for secrets and nothing for conf_keys, the same result as a minimal setup. An empty list is a valid state for both, so the partial-write rollback I built goes away entirely. A failed push logs and is retried on the next manager start. I'd keep it in `gomanager()`: all nodes in a cluster share the storage, and the manager is the one node that can write without racing another.

**3. One-time legacy import.** Agreed on all of it: never modify an existing plugin; import the edge-only plugins that are missing by id (`workflow`, `sshplug`, `sshxplug`, `dockerplug`, `kuberun`, `kubeyaml`, `kubelist`, `stdin`, `terminal`); and for `shellplug` and `testplug`, push a copy with " V2" in the title under a new id, skipping any id already present. `workflow` keeps its id since job.js matches on it. Detection works as you describe: a server_groups record with `master` present and `manager` undefined. Classic seeds `maingrp` with master 1 and `allgrp` with master 0, so I'd copy `manager = master` onto every such group rather than test for 1. Writing `manager` is the fingerprint and happens last, only after the import succeeded. A failed run retries; a completed one never runs again. Classic ignores the extra field, so switching back keeps working. I'd still keep the small read-path alias for `master`: the import has to run from `gomanager()` for the same cluster reason as above, and on the very first boot the node cannot become manager without the alias. Once the fingerprint is written it is never exercised again.

If that matches what you had in mind, I'll rebase onto current main (it applies cleanly) and rework the branch in place, one commit per item so the pieces stay reviewable on their own.

fmaass:
The rework is pushed, and the branch is rebased on current main, one commit per item. Here's what changed from what you proposed:

* The setup walker is gone. `createRequiredLists` checks an explicit array (`global/secrets`, `global/conf_keys`) from `gomanager()` and seeds only the non-optional entries. There is no rollback logic since an empty list is a valid end state.
* `importLegacyPlugins` runs right after it, once, on a folder whose groups have `master` and no `manager`. It imports the missing edge-only plugins by id, adds `shellplug_v2` and `testplug_v2`, and never touches an existing plugin. Only after reading the plugin list back and finding every import present does it copy `manager = master` onto those groups as the fingerprint. A failed import leaves the fingerprint unwritten, so the next manager start retries.
* The lang/theme params are now synthesized by the editor for any plugin with a `script` textarea and removed from the stock plugins in `setup.json`. An off-list stored value falls back to the default. One helper feeds the render, the defaults and `get_event_form_json`. This also closes the hole where injected selects never saved.
* One behavior worth knowing: since `master` stays on the stamped records, the read-path alias lets `manager` decide alone once it is present. Setting `manager` to 0 on such a group actually revokes eligibility.

`npm test` 107 of 107. The PR description is updated to match. I'm happy to adjust the V2 ids or titles if you prefer something else.

mikeTWC1984:
OK, one quick thing. The new pseudo-params are kinda detached from script box (going to the end of the grid). I think we need to stick them together in some compact way.

E.g. 
image

This could be done in get_plugin_params_html.  Quick and dirty example 

```js
						case 'textarea':
							let ta_height = parseInt(param.rows) * 15;
							html += '<div class="plugin_params_label">' + param.title + '</div>';
							html += '<div class="plugin_params_content" style="width: 54rem"><textarea id="fe_ee_pp_' + param.id + '" style="width:99%; height:' + ta_height + 'px; resize:vertical;" spellcheck="false" onkeydown="return catchTab(this,event)">' + escape_text_field_value(value) + '</textarea></div>';
							if (param.id == 'script') {
								html +=`<div style="display: flex; gap: 8px; align-items: center">
								   <div class="plugin_params_label">Syntax:</div>
								   <div><select id="fe_ee_pp_lang"><option value="shell" selected="selected">shell</option><option value="powershell">powershell</option><option value="javascript">javascript</option><option value="python">python</option><option value="perl">perl</option><option value="groovy">groovy</option><option value="java">java</option><option value="csharp">csharp</option><option value="scala">scala</option><option value="sql">sql</option><option value="yaml">yaml</option><option value="toml">toml</option><option value="dockerfile">dockerfile</option><option value="json">json</option><option value="props">props</option></select></div>
								   <div class="plugin_params_label">Theme:</div>
								   <div><select id="fe_ee_pp_theme"><option value="default" selected="selected">default</option><option value="light">light</option><option value="gruvbox-dark">gruvbox-dark</option><option value="solarized light">solarized light</option><option value="solarized dark">solarized dark</option><option value="darcula">darcula</option><option value="ambiance">ambiance</option><option value="base16-dark">base16-dark</option><option value="nord">nord</option></select></div>
								   </div> <br>
								   `

							}

							break;
```

fmaass:
Good call. That placement was left over from rendering the effective list in grid order.

I pushed a change: the two selects now sit in one compact flex row directly under the script box. The textarea case emits the row when the param id is `script`. Both selects are skipped in their old grid position, so nothing appears at the end.

The row reuses the same synthesized params, option lists, and value resolution as before. The ids, CodeMirror binding, and save path are unchanged. An existing plugin record that still declares its own lang/theme params also gets exactly one row.

Checked in the browser on shellplug, a plugin declaring only a script textarea (save round-trip verified in storage), testplug (no row) and a new terminal event (JavaScript preselected).

mikeTWC1984:
OK, looks/work good now. 

Some final polishing:

  I guess we can just keep original Shellplug, Test and URL (http) plugins, lets skip V2 versions. I realized changes are fairly minimal, especially since we have automatic lang/theme generation, so it likely just create confusion.

Also skip SSH and keep SSHX only. I was going to deprecate SSH at some point anyway

mikeTWC1984:
One more optional thing - right now we render lang/theme no matter what  for script
But say user want to use script property to submit some config data to his plugin logic, e.g. sql or yaml file. Would be nice if user can set lang=yaml as hidden parameter, then at render time if we see it's "hidden" we won't render it. Same for theme.

fmaass:
Both done and pushed.

Import: no more V2 copies. The shell, test and http plugins classic ships keep their ids and records, so the import leaves them alone, and the SSH plugin is skipped, so a migrated folder gets SSHX only. The tests follow, and the suite's restore helper now puts back the one stock record the import no longer supplies.

Hidden lang/theme: a plugin that declares `lang` or `theme` as a hidden param pins it. The editor renders no select for that param, the hidden value is saved on every event like any other hidden param, and the script box opens in that mode. If the pinned value is not one of the build's modes, the editor falls back to shell (or the default theme) while the event still saves the plugin's value, since the plugin logic may read it. A hidden `lang` alone leaves the theme select in place, and Terminal's hidden `default_lang` still acts as a changeable default. Checked in a browser with a new event, an event that had stored a different lang under a pinned plugin, both params pinned, and an unsupported value.

One thing I ran into while setting that up: the plugin API rejects any param without a title (`Plugin Parameter Title is missing.`), hidden ones included, while the plugin editor accepts an empty label for a hidden param. So the first save of a hidden `lang` fails until a label is added. I can relax that check for hidden params in this PR if you want, or leave it as is.

mikeTWC1984:
That's ok, let's keep title as  required field.  OK, great work, thank you!

PR #266, opened 2026-08-18
Fix event editor crash when a plugin has no lang/theme params

fmaass:
## Problem

Opening an event's edit view (`#Schedule?sub=edit_event&id=…`) throws when the event's
plugin declares a `script` textarea param but no `lang` and `theme` select params:

```
Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
    at setScriptEditor (combo.min.js:244:6119)
    at gosub_edit_event (combo.min.js:206:27)
    at onActivate (combo.min.js:93:24000)
```

`setScriptEditor()` already allows for the script textarea being absent
(`if (!el) return`), but the two subsequent `addEventListener` calls are unconditional:

```js
document.getElementById("fe_ee_pp_lang").addEventListener("change", …)
document.getElementById("fe_ee_pp_theme").addEventListener("change", …)
```

Those two `<select>` elements are not fixed markup — `get_plugin_params_html()` renders
`fe_ee_pp_<param.id>` directly from the plugin record, so they exist only when the plugin
declares `lang` and `theme` params. Every plugin shipped in `setup.json` does, which is why
the problem is invisible on a fresh install. It appears in two situations:

- a user-defined plugin that has a script textarea but no syntax/theme selectors;
- a `shellplug` record carried over from classic Cronicle, whose params are
  `script` / `annotate` / `json` only.

The exception propagates out of `onActivate`, leaving the single-page app wedged: after
landing on one such event, navigating to a different event's edit view continues to render
the previous one until the page is reloaded.

The same missing `theme` param has a second, quieter effect. In
`let theme = app.getPref('theme') == 'dark' && params.theme == 'default' ? 'gruvbox-dark' : params.theme`
the dark mapping triggers only when `params.theme` is literally `'default'`. With no theme
param, it is `undefined`, the ternary yields `undefined`, and `theme || 'default'`
lands on `cm-s-default` — a white CodeMirror on a dark page.

## Change

Three lines in `setScriptEditor()`:

- attach the syntax listener only if `#fe_ee_pp_lang` exists;
- attach the theme listener only if `#fe_ee_pp_theme` exists;
- treat an unset `params.theme` the same as `'default'` when resolving the dark theme.

I kept the guards rather than emitting the selects unconditionally: a plugin without `lang`
and `theme` params is a legitimate shape, and forcing the elements would require writing
those params into every plugin record. Guarding also matches what the function already does
for the script textarea one screen above.

## Verification

I built images from `main` and this branch, ran both against the *same* data directory
(so only the image differed), and drove a real Chromium against the UI in dark mode.

Two events on a fresh install, both using a shell script:

- control — stock `shellplug` (has `lang` + `theme`);
- repro — a plugin declaring `script` / `annotate` / `json` only.

| | `main` | this branch |
|---|---|---|
| control: CodeMirror class | `cm-s-gruvbox-dark` | `cm-s-gruvbox-dark` |
| control: page errors | 0 | 0 |
| repro: CodeMirror class | `cm-s-default` (background `rgb(255,255,255)`) | `cm-s-gruvbox-dark` (background `rgb(40,40,40)`) |
| repro: page errors | 1 — the `addEventListener` TypeError | 0 |
| navigate repro → control (hash nav) | still shows the repro event | shows the control event |

Regressions I checked on the built branch image:

- light mode is unchanged — both events resolve to `cm-s-default`;
- the syntax select still switches the editor mode (`shell` → `python`) on a plugin that has it;
- the theme select still switches the editor theme (`gruvbox-dark` → `nord`) on a plugin that has it;
- the guards do not fabricate elements — `#fe_ee_pp_lang` / `#fe_ee_pp_theme` are still absent
  on the repro event;
- editing the script still writes back into the underlying textarea.

I confirmed that the change survives bundling by grepping the built `combo.min.js` in both images:
the unguarded `getElementById("fe_ee_pp_lang").addEventListener` form appears once on
`main` and is absent from this branch, replaced by the guarded assignment.

I did not run the automated test suite; the change has no server-side surface.

### DIFF
--- a/htdocs/js/pages/Schedule.class.js
+++ b/htdocs/js/pages/Schedule.class.js
@@ -2918,7 +2918,8 @@ Class.subclass(Page.Base, "Page.Schedule", {
 		}
 		if (lang == 'props') { lang = 'text/x-properties' }
 
-		let theme = app.getPref('theme') == 'dark' && params.theme == 'default' ? 'gruvbox-dark' : params.theme;
+		// a plugin without a "theme" param leaves params.theme unset - treat that as 'default'
+		let theme = app.getPref('theme') == 'dark' && (!params.theme || params.theme == 'default') ? 'gruvbox-dark' : params.theme;
 		if (params.theme == 'light') theme = 'default'
 
 		let editor = CodeMirror.fromTextArea(el, {
@@ -2941,8 +2942,9 @@ Class.subclass(Page.Base, "Page.Schedule", {
 
 		editor.on('change', (cm) => { el.value = cm.getValue() })
 
-		// syntax selector
-		document.getElementById("fe_ee_pp_lang").addEventListener("change", function () {
+		// syntax selector - only rendered if the plugin declares a "lang" param
+		let langSelect = document.getElementById("fe_ee_pp_lang")
+		if (langSelect) langSelect.addEventListener("change", function () {
 			let ln = this.options[this.selectedIndex].value;
 
 			editor.setOption("gutters", ['']);
@@ -2967,8 +2969,9 @@ Class.subclass(Page.Base, "Page.Schedule", {
 			editor.setOption("mode", ln);
 		});
 
-		// theme 
-		document.getElementById("fe_ee_pp_theme").addEventListener("change", function () {
+		// theme - only rendered if the plugin declares a "theme" param
+		let themeSelect = document.getElementById("fe_ee_pp_theme")
+		if (themeSelect) themeSelect.addEventListener("change", function () {
 			var thm = this.options[this.selectedIndex].value;
 			if (thm === 'default' && app.getPref('theme') === 'dark') thm = 'gruvbox-dark';
 			if (thm === 'light') thm = 'default';
--- a/htdocs/js/pages/Schedule.class.js
+++ b/htdocs/js/pages/Schedule.class.js
@@ -2896,7 +2896,10 @@ Class.subclass(Page.Base, "Page.Schedule", {
 		let privs = app.user.privileges;
 		let canEdit = privs.admin || privs.edit_events || privs.create_events;
 
-		let lang = params.lang || params.default_lang || 'shell';
+		// the param grid renders before this runs and always pairs a script control with both
+		// selects, so they hold the fallback for an event that stored no syntax/theme of its own.
+		// a plugin may declare lang/theme as some other input type - ignore those and fall through
+		let lang = params.lang || params.default_lang || (document.querySelector('select#fe_ee_pp_lang') || {}).value || 'shell';
 		// gutter for yaml linting
 		let gutter = ''
 		let lint = 'false'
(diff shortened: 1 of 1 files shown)


htdocs/js/pages/Schedule.class.js
```diff
@@ -2918,7 +2918,8 @@ Class.subclass(Page.Base, "Page.Schedule", {
 		}
 		if (lang == 'props') { lang = 'text/x-properties' }
 
-		let theme = app.getPref('theme') == 'dark' && params.theme == 'default' ? 'gruvbox-dark' : params.theme;
+		// a plugin without a "theme" param leaves params.theme unset - treat that as 'default'
```
mikeTWC1984: This comment describes empty theme handling, not what next line is doing.

htdocs/js/pages/Schedule.class.js
```diff
@@ -2941,8 +2942,9 @@ Class.subclass(Page.Base, "Page.Schedule", {
 
 		editor.on('change', (cm) => { el.value = cm.getValue() })
 
-		// syntax selector
-		document.getElementById("fe_ee_pp_lang").addEventListener("change", function () {
+		// syntax selector - only rendered if the plugin declares a "lang" param
+		let langSelect = document.getElementById("fe_ee_pp_lang")
```
mikeTWC1984: We need to enforce lang and theme select boxes to render, rather than ignore them - otherwise you'd loose functionality to select it (I guess) Need to assign default values when you render parameter grid (somewhere in Schedule.class.js) , not here.

fmaass:
Both reworked in 812283b.

Line 2921: the comment now describes the line right under it.

The main change: defaults are assigned where the parameter grid renders, in `get_plugin_params_html`. When a plugin renders a script control without declaring lang/theme, both selects are appended to a local copy of `plugin.params` (`plugin.params` itself is never mutated), defaulted like the stock plugins — shell / default. A value already stored on the event still wins.

The injection keys off the param *type*, not just the id: only text/textarea/checkbox/select emit an `fe_ee_pp_<id>` element, so a hidden-typed lang param — which renders nothing — still gets a select supplied and shows the pinned value. The trigger mirrors `setScriptEditor`'s own entry gate exactly, so every shape that reaches the editor has both selects, and the addEventListener lines are back to upstream verbatim — no null guards, the crash no longer occurs structurally. The editor's syntax/theme fallback reads select elements specifically, so a lang declared as some other input type falls through to the stock defaults instead of adopting an unrelated control's value.

The selects stay editor-local: `get_event_form_json` still builds `event.params` from the plugin's own param list, so nothing a plugin never declared is written back — I saved such an event and read it back through the API to check.

Verified in a browser against a container built off the branch: plugins with and without lang/theme, a hidden-pinned lang and a checkbox-typed lang all open with zero JS errors and working selects; without the fix the original TypeError reproduces. npm test green.

fmaass:

PR #268, opened 2026-08-19
fix: add output encoding in APIKeys.js

anupamme:
## Summary

The admin API-key list currently renders the complete API key in the page DOM. This unnecessarily exposes a credential to any code or tooling with access to the rendered page and makes accidental disclosure easier.

This change masks the key in the admin list while preserving the existing API-key functionality. The first four characters are retained to allow administrators to distinguish keys.

This is a defense-in-depth credential exposure reduction rather than an assertion that the existing \`encode_entities()\` usage is vulnerable to HTML injection.

## Changes

- `htdocs/js/pages/admin/APIKeys.js` — normalize `item.key` with `String(item.key || '')` before calling `.substring()`, so the list renders safely even when a key value is null, undefined, or a non-string type; then mask all but the first 4 characters.

### DIFF
--- a/htdocs/js/pages/admin/APIKeys.js
+++ b/htdocs/js/pages/admin/APIKeys.js
@@ -56,7 +56,7 @@ Class.add( Page.Admin, {
 			];
 			return [
 				'<div class="td_big">' + self.getNiceAPIKey(item, true, col_width) + '</div>',
-				'<div style="">' + encode_entities(item.key) + '</div>',
+				'<div style="">' + encode_entities(item.key.substring(0, 4) + '********************************') + '</div>',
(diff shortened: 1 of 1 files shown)


mikeTWC1984:
Fair point. I'd convert it to string before calling substirng as a sanity check (like if it's  null or object - it will crash UI)

anupamme:
Thanks, good catch. I've updated the code to normalize `item.key` with `String(item.key || '')` before calling `substring()` so null or non-string values can't break the admin UI.

I've also tightened the PR description to frame this as reducing unnecessary API-key exposure in the admin DOM rather than as an HTML/XSS injection issue. The intent is to keep the change narrowly scoped to masking the credential while preserving the existing UI behavior.

PR #269, opened 2026-08-19
fix(auth): recover expired proxy sessions

frankii91:
## Summary

- reload the current page once when `user/resume_session` receives an HTTP `401`
- keep the existing application and network error behavior for every other response
- use a tab-scoped 30-second guard to prevent a persistent proxy rejection from causing a reload loop
- add focused regression coverage for reload, loop prevention, guard cleanup, non-401 errors, and unavailable browser storage

## Why

External authentication proxies can reject an AJAX request after their own browser session expires, before the request reaches Cronicle. Cloudflare Access documents this behavior for requests carrying `X-Requested-With: XMLHttpRequest` and recommends either forcing a page refresh or displaying an expired-session message:

https://developers.cloudflare.com/cloudflare-one/access-controls/access-settings/session-management/#ajax

Cronicle's same-origin jQuery requests carry that header. The existing `resume_session` call has no failure recovery, so the browser can remain on a partially rendered application shell while a normal top-level navigation would allow the proxy to start its interactive login flow.

The API client reports HTTP statuses as numeric error codes, while Cronicle application errors use string codes. The new path therefore activates only for numeric `401`, reloads the same URL without accepting a redirect target from the response, and leaves Cronicle's normal missing or expired local-session flow unchanged. No deployment-specific configuration is required.

The reload timestamp is stored only in the current tab's `sessionStorage`. It is cleared after a successful `resume_session`; a second `401` within 30 seconds or unavailable storage falls back to the existing visible network error instead of reloading again.

## Tests

- before the implementation, the focused regression failed because `resume_session` registered no error callback
- `npx pixl-unit test/session-recovery.test.js`: `1/1` passed
- `npm test`: `92/92` passed (`374` assertions)
- `node --check htdocs/js/pages/Base.class.js`
- `node --check test/session-recovery.test.js`
- `git diff --check`

### DIFF
--- a/htdocs/js/pages/Base.class.js
+++ b/htdocs/js/pages/Base.class.js
@@ -5,6 +5,17 @@ Class.subclass(Page, "Page.Base", {
 	requireLogin: function (args) {
 		// user must be logged into to continue
 		var self = this;
+		var session_reload_key = 'cronicle_resume_session_reload:' + window.location.pathname;
+		var session_reload_limit = 30000;
+
+		function show_session_error(resp) {
+			if (resp && (typeof(resp.code) == 'number')) {
+				app.doError("Network Error: " + resp.code + ": " + resp.description);
+			}
+			else {
+				app.doError("Error: " + ((resp && resp.description) || "Unknown error"));
+			}
+		}
 
 		if (!app.user) {
 			// require login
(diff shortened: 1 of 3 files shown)


mikeTWC1984:
OK, looks good. Some cosmetic corrections - for that second error handling function, can you use "err" vs "resp". It also not immediately obvious that error handler is argument and not part of the success function.  Those are minor things but figuring out what is what took most of my reviewing time :)

app.api.post('user/resume_session', {session_id: session_id}
  // success response
  ,function (resp) {
  }
 // error response
 , function(err) {
 });

PR #280, opened 2026-08-20
build: pin active container base images

frankii91:
## Problem

The active container build consumed mutable base-image references, so the same source revision could resolve to different build inputs over time.

## Change

- Pin both stages of the active root `Dockerfile` to the same immutable multi-architecture `cronicle/base-alpine` index digest.
- Pin the Alpine builder definition to an immutable `alpine:3.22.2` index digest.
- Keep the readable tags alongside the digests.

This is intentionally scoped to the active publication path.  The unused legacy `Docker/arm64.dockerfile` is not presented as covered by this PR.

## Validation

- Verified both index digests include the required amd64, arm64, and arm/v7 platforms.
- Both `docker buildx --check` runs pass without warnings.
- Real no-cache builds completed for the application image and helper base; runtime smoke confirmed Node, tini, manager/worker entrypoints, and expected data/config permissions.
- Full `npm test`: 92/92 tests, 375 assertions.
- CRLF-aware `git diff --check`: pass.

The issue and initial patch were proposed by a Codex Ultra security review. This version was manually re-analyzed against the active multi-architecture build path, validated with real builds and runtime smoke tests, and is submitted for maintainer review in line with our prior discussion.

### DIFF
--- a/Docker/alpine-base.dockerfile
+++ b/Docker/alpine-base.dockerfile
@@ -6,6 +6,6 @@
 # multi-arch build
 # docker buildx build --platform linux/amd64,linux/arm64,linux/arm/v7 -t cronicle/base-alpine:v3.22.2 --push -f Docker/alpine-base.dockerfile .
 
-FROM alpine:3.22.2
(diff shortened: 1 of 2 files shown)


mikeTWC1984:
I guess it make sense. I think it's time to bump alpine version to something newer, I'll update and include shas when publishing release

PR #281, opened 2026-08-20
fix: scrub AWS session credentials

frankii91:
## What changes

- hide camel-case AWS `sessionToken` values in Config Viewer
- remove legacy and nested AWS `secretAccessKey` / `sessionToken` environment variables during startup
- keep the regression coverage in the existing `lib/test.js` suite

## Concrete behavior

For example, when Cronicle starts with `CRONICLE_Storage__AWS__sessionToken`, the value is first loaded into server configuration. Before this change, a plugin process could still inherit that environment variable and Config Viewer could return the token unchanged.

After this change, the loaded server configuration keeps the credential for the storage engine, but the startup environment no longer contains it and Config Viewer returns `[REDACTED]`. Non-secret AWS fields such as `region` and `accessKeyId` remain visible.

Both the legacy direct AWS layout and the nested `credentials` layout are covered.

## Validation

- `npm ci`
- `npm test` — 97/97 tests, 448 assertions, 2 suites
- `node --check lib/engine.js`
- `node --check lib/api/admin.js`
- `node --check lib/test.js`
- `node --check test/oidc-logout.test.js`
- `git diff --check upstream/main...HEAD`

### DIFF
--- a/lib/api/admin.js
+++ b/lib/api/admin.js
@@ -379,7 +379,8 @@ module.exports = Class.create({
 		const sensitiveKeys = new Set([
 			'secret', 'token', 'secret_key', 'client_secret', 'password', 'passwd', 'private_key',
 			'access_token', 'refresh_token', 'id_token', 'id_token_hint',
-			'oidc_id_token_hint', 'oidc_id_token_hint_enc', 'logout_token', 'api_key', 'secretaccesskey'
+			'oidc_id_token_hint', 'oidc_id_token_hint_enc', 'logout_token', 'api_key',
+			'secretaccesskey', 'sessiontoken'
 		]);
 		const redact = function(value) {
 			if (!value || typeof value !== 'object') return;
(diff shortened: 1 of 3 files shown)


mikeTWC1984:
Looks OK. The only thing the way you are adding test seem a bit cumbersome. Like for every small thing you have separate file with long name, then add this file under npm test section in package.json

pixl-unit should support globs I believe (check docs), so I'd rather keep npm test static omething like "pixl-unit lib/test/.*js" , see if you can make it work.

mikeTWC1984:
It now conflicts  with https://github.com/cronicle-edge/cronicle-edge/pull/270 , which a bit more cleaner to my mind. I'd do similar approach to this PR to - move the test to test.js rather crating new file, and do inline delete process.env to resolve conflict. Can do scrubber function sometime later (if adding new item to scrub), but to me personally explicit delete is a easier to follow

frankii91:
Updated as suggested: rebased on current `main` (including #270), removed the scrubber and separate test file, kept the explicit `delete process.env[...]` style, and moved the coverage into `lib/test.js`.

Local tests pass (97/97, 448 assertions), and both GitHub Actions test jobs are green.

PR #283, opened 2026-08-20
fix: fail closed on deprecated git bootstrap

frankii91:
## Summary

- fail closed in both Unix and Windows manager entrypoints when deprecated `GIT_REPO` bootstrap is configured
- never print the configured repository value, which may contain credentials
- propagate setup failures so a later `start`/manager runtime cannot run against incomplete storage
- update the Git integration documentation to match the currently disabled/deprecated implementation and point to explicit import workflows

## Why

The repository-clone block in the current Unix manager is commented out and has been deprecated for years, but the README still advertises `GIT_REPO` as a supported Docker restore path. Re-enabling that old block does not provide a safe restore: a partial checkout can still fall through to setup and seed default state.

This change deliberately does **not** reactivate automatic restore or add persistent recovery state. It makes the unsupported configuration explicit and ensures a failed `storage-cli.js setup` stops before the runtime on both manager entrypoints.

## Validation

- focused entrypoint suite — 9/9
- full `npm test` — 101/101 tests, 375 assertions, 3 suites
- local no-network container/entrypoint smoke — 9/9
- `bash -n bin/manager bin/control.sh`
- `dash -n bin/control.sh`
- static Windows `cmd.exe` source-contract regression for the guard and both setup paths
- `node --check test/bootstrap-entrypoints.test.js`
- `git diff --check`

The Windows batch file was not executed under a real Windows `cmd.exe` in this environment; its control-flow contract is covered statically.

## Review provenance

The issue and an initial patch were proposed by a Codex Ultra security review (fork PR [frankii91/cronicle-edge#19](https://github.com/frankii91/cronicle-edge/pull/19)). That generated patch was not ported because it reactivated the deprecated clone path and treated incomplete storage as restored. This fail-closed version was manually re-analyzed across Unix/Windows bootstrap and setup lifecycles, independently reviewed, and is submitted for maintainer review in line with our prior discussion.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -183,8 +183,8 @@ You can generate a token to run specific event instead of creating global API ke
 ### Git integration
  - Github - HMAC signature check is supported. You can set `git_hub_key` config to verify signature. If not set cronicle will use `secret_key` (if x-signature-header is included in the request).
  - Gitlab - you can use cronicle API key as web hook secret key, but you can always bake API key or token in web hook url.
- - You can also automatically execute "git add / git commit / git push remote branch" on clicking backup button or even on each metadata update. First you need to set git repo (with auth) in your data folder. Then use related configs (there is an example in demo configs).
- - Before starting new Docker container you can also set `GIT_REPO` variable, so manager entrypoint will attempt to use this repo to setup data folder instead of running standard set up.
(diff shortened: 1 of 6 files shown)


mikeTWC1984:
Just removing GIT_REPO info from readme would be sufficient. I'll do it later

PR #286, opened 2026-08-20
fix: scope workflow capabilities to their parent job

frankii91:
## Summary

- bind workflow authentication to a live or queued parent workflow job instead of a manager-local bearer registry;
- derive the effective child-event allowlist from the parent job's immutable workflow snapshot, including disabled/start-step handling;
- preserve supported cross-category workflows only for child IDs explicitly present in that snapshot;
- allow workflow credentials on exactly five canonical API routes, with regression coverage for path-confusion variants;
- restrict aborts to jobs whose `source_id` is the authenticated parent and deny bulk aborts.

## Security impact

**High, deployment-dependent.** A valid workflow capability on current `main` can cross event RBAC boundaries and invoke or abort configured events outside the caller's normal category/group scope. The rating is High rather than universally Critical because exploitation still requires a valid capability for an active workflow, the Workflow plugin, and an impactful configured child event; it does not grant administrator, edit, or arbitrary API privileges.

## Dependencies and compatibility

This branch currently includes the still-open changes from #257 and #284. Those PRs should merge first; this branch will then be rebased so only the workflow-capability commit remains.

The existing workflow signature wire format is retained for rolling compatibility, but comparison is timing-safe and the signature is recomputed from the current parent job. No durable bearer registry is added.

Authentication can be reconstructed from an authenticated remote parent snapshot once a request reaches the current manager. Existing workflow-child transport remains pinned to the launch manager's `BASE_URL` and does not follow manager redirects, so end-to-end continuity of an already running workflow after manager failover is explicitly outside this PR.

## Validation

- full `npm test`: 100/100, 745 assertions;
- focused workflow-capability suite: 8/8, 176 assertions;
- real HTTP tests for canonical allowed routes, denied routes, ownership, malformed routes, suffix/path-confusion routes, and active/pending/retry parent snapshots;
- `node --check` and `git diff --check`;
- independent adversarial review: approved.

Initially surfaced by Codex Security / Codex Ultra, then manually re-analyzed, corrected, regression-tested, and independently reviewed. Submitted for maintainer review.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -525,6 +525,171 @@ module.exports = Class.create({
 
 	},
 
+	isSafeJobIdentifier: function(id) {
+		return (typeof id == 'string') && /^\w+$/.test(id) &&
+			[ '__proto__', 'prototype', 'constructor' ].indexOf(id) < 0;
+	},
+
+	isSafeEventIdentifier: function(id) {
+		return (typeof id == 'string') && /^\w+$/.test(id);
+	},
+
+	getWorkflowSignature: function(job_id) {
+		// Keep the existing wire format for rolling upgrades.  The signature is
+		// derived when needed, rather than retained in a manager-local bearer map.
+		return Tools.digestHex(job_id + this.server.config.get('secret_key'), 'MD5');
+	},
+
+	workflowSignatureMatches: function(job_id, supplied) {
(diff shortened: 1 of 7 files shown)


mikeTWC1984:
I think I'll create a discussion for this topic. We need to go over general design of how WF work and what flaws this is addressing.

PR #287, opened 2026-08-21
fix: bind worker run-as identity to signed job dispatch

frankii91:
## Summary

This change preserves administrator-controlled Plugin UID/GID across manager-to-worker dispatch and makes the one-shot admin `debug_sudo` path safe across mixed-version clusters.

It:

- serializes Plugin UID/GID independently of the manager operating system;
- preserves numeric `0` and string `"0"` without truthiness fallbacks;
- removes trust in legacy top-level `job.debug_sudo`;
- introduces a versioned, HMAC-SHA256-authenticated run-as dispatch context;
- binds it to a canonical digest of the complete job payload, plus the job, manager, worker, socket session, sequence, mode, Plugin, UID, and GID;
- prevents replay within or across worker connections;
- consumes the context before active, pending, retry, recovery, or completed state;
- keeps capacity-queued jobs fail-closed under their configured Plugin identity.

There is no new persistent schema or durable authorization state.

## Security impact and severity

**Proposed severity: High**

The manager selected or serialized job identity, while the worker ultimately performed the Unix process spawn. A Windows manager did not serialize Plugin UID/GID because identity handling was conditional on the manager platform. A Unix worker could consequently fall back to the Cronicle service UID, which is commonly `root`. Separately, a worker could not distinguish an authorized one-shot admin `debug_sudo` request from a legacy field preserved by an older manager or job record.

An authenticated non-admin caller with `run_events` access to an affected event could therefore cause its configured Plugin command to run under the worker service identity instead of the administrator-configured restricted identity. For a shell-like or otherwise attacker-influenceable Plugin, this can become service-level or root command execution.

The highest-impact path requires:

1. an authenticated caller with applicable `run_events`, category, and group access;
2. an event using a Plugin configured with a restricted UID/GID, or a legacy job carrying `debug_sudo`;
3. a Unix worker whose Cronicle service identity is more privileged than the Plugin identity;
4. commonly, a Windows-manager-to-Unix-worker topology;
5. a Plugin whose command or parameters make the identity change security-relevant.

This is rated High rather than Critical because it is not an unauthenticated default remote compromise: the caller already needs scoped event-execution authority, the affected topology or legacy-marker condition must exist, root impact requires a root-running worker service, and useful command execution depends on the configured Plugin behavior. The impact can nevertheless reach root on an affected worker, so Medium would understate it.

## Signed dispatch boundary

New managers advertise `signed_job_run_as_v2`. A capable worker acknowledges it and returns a fresh random 256-bit dispatch-session identifier. The earlier v1 draft was never published and is intentionally rejected rather than treated as a compatibility protocol.

For every remote job sent to a v2-capable worker, the manager computes a canonical SHA-256 digest of the complete JSON-normalized wire job, excluding only the top-level `_cronicle_run_as` context. It then creates a transport-only context containing the protocol version, `plugin` or `service` mode, job ID, manager hostname, worker hostname, dispatch session, monotonically increasing sequence, payload digest, and HMAC-SHA256 signature. The signature uses the existing cluster secret with domain-separated context and the canonical JSON payload digest, so numeric and string UID/GID values cannot be confused and the authorization cannot be detached from the exact command, arguments, environment, files, workflow data, or other execution inputs.

The worker recomputes the canonical payload digest and verifies the context and sequence before `launchLocalJob`. The context cannot be rebound to another payload, job, manager, worker, or connection, and a consumed sequence cannot be replayed. Cycles, unsupported values, and other non-canonical payloads fail closed. For `service` mode, the worker materializes its own local service UID. It then removes the signed context and any legacy `debug_sudo` field before queueing, persistence, recovery, retry, or completed storage.

## Rolling upgrade order

This protocol intentionally requires a **manager-first** rolling upgrade.

1. Upgrade managers first. A new manager gates dispatch during negotiation. An old worker receives no `_cronicle_run_as` or `debug_sudo` marker, preserves Plugin UID/GID, and safely degrades an admin debug request to the Plugin identity.
2. Drain and upgrade workers. A new worker rejects authentication from an old manager that does not advertise signed dispatch. Worker-first deployment may therefore make that worker temporarily unavailable, but cannot silently elevate a job.
3. New manager and new worker negotiate a fresh session and use signed, replay-protected dispatch for every remote job.

Jobs already active, pending, retrying, or recoverable inside an old worker remain governed by that old binary until drained or completed. This PR deliberately does not rewrite old worker-owned runtime state. Operators should drain existing work and upgrade workers promptly after the manager upgrade.

## Persistent-state boundary

The authorization lifecycle is intentionally short:

1. `run_event` records an admin debug request only in an in-memory launch option.
2. The manager's canonical job contains neither `debug_sudo` nor `_cronicle_run_as`.
3. A capable remote worker receives the signed context only in a deep-copied wire payload.
4. The worker verifies and consumes it synchronously.
5. Active jobs, delayed jobs, retry jobs, recovery JSON, completed jobs, and persistent event queues contain no dispatch context or debug marker.
6. Capacity deferral drops debug authority and retries under the normal Plugin identity.

Only the negotiated session and sequence are ephemeral socket/worker connection state. They are cleared on disconnect and never written to storage.

## Dependencies and tested DAG

This change is based on the published head of [#286](https://github.com/cronicle-edge/cronicle-edge/pull/286), which includes the relevant work from [#284](https://github.com/cronicle-edge/cronicle-edge/pull/284), [#257](https://github.com/cronicle-edge/cronicle-edge/pull/257), and the corrected workflow-capability scope.

```text
upstream/main 9ebbaf3
  -> #284 12989d9
  -> #257-equivalent caa8fe7
  -> #286 78c2695
  -> this change
```

This PR should be reviewed and merged after #286, or rebased onto `main` after #286 lands.

## File scope

Exactly seven files are changed relative to #286:

- `htdocs/js/pages/admin/Plugins.js`: preserves numeric/string UID/GID, including zero.
- `lib/api.js`: strips run-as transport fields at the event persistence boundary.
- `lib/api/event.js`: converts an admin `debug_sudo` request into a one-shot launch option.
- `lib/api/plugin.js`: accepts validated numeric or string UID/GID values.
- `lib/comm.js`: negotiates capability/session/sequence, gates dispatch, and rejects failed verification.
- `lib/job.js`: handles platform-neutral Plugin identity, computes/verifies the signed payload context, and consumes worker-local run-as state.
- `lib/test.js`: adds authorization, mixed-version, replay, prototype, zero-value, queue, and lifecycle regressions.

No `lib/api/job.js` or `lib/engine.js` change is introduced by this PR relative to #286.

## Validation

The exact patch was integrated on #286 head `78c269516d3767760959b8463e0b2795e8abc3be`.

```text
Tests passed: 104 of 104
Tests failed: 0
Assertions: 910
Test suites: 2
```

`git diff --cached --check` and `node --check` passed for all changed JavaScript files.

A focused test with real Socket.IO server/client connections on random loopback ports covered:

| Manager | Worker | Result |
|---|---|---|
| old | new | authentication rejected; no launch |
| new | old | clean legacy payload; debug degrades to Plugin UID/GID |
| new | new | signed context accepted; worker-local service UID applied only after verification |
| replayed new-manager packet | new | repeated sequence rejected and connection authorization revoked |

Regression coverage also includes Windows-manager-to-Unix-worker serialization, numeric and string zero, HMAC tampering, mutation of command, working directory, parameters, environment, loader options, stdin, files, workflow and chain data, wrong job/manager/worker/session, stale ACK and reconnect state, inherited/prototype-polluted fields, non-admin create/update/run attempts, and active/pending/retry/recovery/capacity/completed lifecycles.

This change secures the run-as authorization boundary. During the short legacy-worker phase, the pre-existing plaintext cluster channel still does not provide complete payload integrity, and non-launch cluster controls retain their existing transport properties. Those broader transport concerns are outside this PR. The required manager-first order and prompt worker drain keep the corrected `debug_sudo` boundary fail-closed throughout the supported rollout.

## Provenance and review disclosure

The original finding and generated proposal [frankii91/cronicle-edge#42](https://github.com/frankii91/cronicle-edge/pull/42) came from Codex Security / Codex Ultra. It was not ported 1:1: independent review demonstrated that its top-level `job.debug_sudo` marker was unsafe across mixed versions. That design was discarded and replaced with the versioned, HMAC-bound, capability-negotiated protocol above.

Initially surfaced by Codex Security / Codex Ultra, then manually re-analyzed, corrected, regression-tested, and independently reviewed. Submitted for maintainer review.

### DIFF
--- a/htdocs/js/pages/admin/Plugins.js
+++ b/htdocs/js/pages/admin/Plugins.js
@@ -187,8 +187,8 @@ Class.add( Page.Admin, {
 				newPlugin.ipc = !!plugin.ipc
 				newPlugin.wf = !!plugin.wf
 				newPlugin.stdin = !!plugin.stdin
-				if(typeof plugin.uid === 'string' || parseInt(plugin.uid)) newPlugin.uid = plugin.uid
-				if(typeof plugin.gid === 'string' || parseInt(plugin.gid)) newPlugin.gid = plugin.gid
+				if(typeof plugin.uid === 'string' || typeof plugin.uid === 'number') newPlugin.uid = plugin.uid
+				if(typeof plugin.gid === 'string' || typeof plugin.gid === 'number') newPlugin.gid = plugin.gid
 				if(typeof plugin.cwd === 'string') newPlugin.cwd = plugin.cwd
 				if(typeof plugin.script === 'string') newPlugin.script = plugin.script 
 
@@ -564,7 +564,7 @@ Class.add( Page.Admin, {
 		html += get_form_table_spacer();
 		
 		// advanced options
-		var adv_expanded = !!(plugin.cwd || plugin.uid);
+		var adv_expanded = !!(plugin.cwd || plugin.uid || plugin.gid || (plugin.uid === 0) || (plugin.gid === 0));
 		html += get_form_table_row( 'Advanced', 
 		`<div autocomplete="off" style="font-size:13px;${adv_expanded ? 'display:none;' : ''}"><span class="link addme" onMouseUp="$P().expand_fieldset($(this))"><i class="fa fa-plus-square-o">&nbsp;</i>Advanced Options</span></div>
 		<fieldset style="padding:10px 10px 0 10px; margin-bottom:5px;${adv_expanded ? '' : 'display:none;'}"><legend class="link addme" onMouseUp="$P().collapse_fieldset($(this))"><i class="fa fa-minus-square-o">&nbsp;</i>Advanced Options</legend>
--- a/lib/api.js
+++ b/lib/api.js
@@ -264,6 +264,8 @@ module.exports = Class.create({
 		delete event.uid;
 		delete event.gid;
 		delete event.env;
+		delete event.debug_sudo;
+		delete event._cronicle_run_as;
 
 
 		// make sure title doesn't contain HTML metacharacters
--- a/lib/api.js
+++ b/lib/api.js
@@ -525,6 +525,171 @@ module.exports = Class.create({
 
 	},
 
+	isSafeJobIdentifier: function(id) {
+		return (typeof id == 'string') && /^\w+$/.test(id) &&
+			[ '__proto__', 'prototype', 'constructor' ].indexOf(id) < 0;
+	},
+
+	isSafeEventIdentifier: function(id) {
+		return (typeof id == 'string') && /^\w+$/.test(id);
+	},
+
+	getWorkflowSignature: function(job_id) {
+		// Keep the existing wire format for rolling upgrades.  The signature is
+		// derived when needed, rather than retained in a manager-local bearer map.
+		return Tools.digestHex(job_id + this.server.config.get('secret_key'), 'MD5');
+	},
+
+	workflowSignatureMatches: function(job_id, supplied) {
+		if (!this.isSafeJobIdentifier(job_id) || (typeof supplied != 'string')) return false;
+		var expected = Buffer.from(this.getWorkflowSignature(job_id), 'utf8');
+		var actual = Buffer.from(supplied, 'utf8');
+		if (actual.length != expected.length) return false;
+		return crypto.timingSafeEqual(expected, actual);
+	},
+
+	isTrustedWorkflowJob: function(job, job_id) {
+		if (!job || (typeof job != 'object')) return false;
+		if (!Object.prototype.hasOwnProperty.call(job, 'id') || (job.id !== job_id)) return false;
+
+		var is_builtin = Object.prototype.hasOwnProperty.call(job, 'plugin') && (job.plugin === 'workflow');
+		var is_custom = Object.prototype.hasOwnProperty.call(job, 'is_wf') && !!job.is_wf;
+		return is_builtin || is_custom;
+	},
+
+	getWorkflowEventMap: function(job) {
+		// Authorize against the immutable workflow snapshot carried by the parent
+		// job.  This deliberately supports children in other categories / targets.
+		var allowed = Object.create(null);
+		var workflow = (job && Object.prototype.hasOwnProperty.call(job, 'workflow') && Array.isArray(job.workflow)) ?
+			job.workflow : [];
+		var options = (job && Object.prototype.hasOwnProperty.call(job, 'options') && job.options &&
+			(typeof job.options == 'object') && !Array.isArray(job.options)) ?
+			job.options : {};
+		var raw_start = options.wf_start_from_step || 1;
+		if (!raw_start) raw_start = 1;
+		var start_from;
+		try { start_from = Number(raw_start); }
+		catch (err) { return allowed; }
+		if (!Number.isFinite(start_from)) return allowed;
+
+		for (var idx = 0; idx < workflow.length; idx++) {
+			if (!Object.prototype.hasOwnProperty.call(workflow, idx)) continue;
+			var step = workflow[idx];
+			if (!step || (typeof step != 'object') || Array.isArray(step)) continue;
+			if ((idx + 1) < start_from) continue;
+			if (step.disabled) continue;
+			if (!Object.prototype.hasOwnProperty.call(step, 'id')) continue;
+			if (!this.isSafeEventIdentifier(step.id)) continue;
+			allowed[step.id] = true;
+		}
+
+		return allowed;
+	},
+
+	getWorkflowCapability: function(job_id, signature) {
+		if (!this.isSafeJobIdentifier(job_id)) return false;
+		var job = this.findJob(job_id);
+		if (!this.isTrustedWorkflowJob(job, job_id)) return false;
+		if (!this.workflowSignatureMatches(job_id, signature)) return false;
+		return { job: job, event_ids: this.getWorkflowEventMap(job) };
+	},
+
+	workflowCanRunEvent: function(user, event_id) {
+		if (!user || !this.isSafeJobIdentifier(user.workflow) || !this.isSafeEventIdentifier(event_id)) return false;
+		var parent = this.findJob(user.workflow);
+		if (!this.isTrustedWorkflowJob(parent, user.workflow)) return false;
+		var event_ids = this.getWorkflowEventMap(parent);
+		return Object.prototype.hasOwnProperty.call(event_ids, event_id);
+	},
+
+	workflowIsActive: function(user) {
+		if (!user || !this.isSafeJobIdentifier(user.workflow)) return false;
+		return this.isTrustedWorkflowJob(this.findJob(user.workflow), user.workflow);
+	},
+
+	workflowOwnsJob: function(user, job) {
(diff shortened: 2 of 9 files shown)


mikeTWC1984:
I don't really understand that problem. 
So it's saying that if manager is Windows and worker is unix, non-admin user can use sudo option? Can AI provide any concrete example how to achieve that?
Generally those PR description are very abstract / AI-sloppish. For small commits I usually just figure out what it does form code change. But here I don't even understand what is it doing with uid/guid. 

Suggestion for AI generate commits:
 - ask it provide concrete example. E.g. it adds some extra parsing of uid/gid. I need an example of uid/gid values that can break current version
 - Add a prompt "assume reviewers is 5 years old" :). Generally I don't understand those descriptions

mikeTWC1984:
It also seemed to be another "super" PR, or it all related? You can combine few small unrelated things together in 1 PR, that's fine. But if a feature requires 300 lines of code on core functions - it should go separate

mikeTWC1984:
Let's hold done on new PRs for now. Let me digest current queue :))

frankii91:
You are right — the description bundled too much together. The shortest concrete example is:

### Concrete security example

**Setup**

- Manager: Windows
- Worker: Linux, with the Cronicle service running as `root` (UID/GID `0`)
- An administrator configures the Shell Plugin with:
  - Run as user: `nobody`
  - Run as group: `nogroup`
  - (commonly UID/GID `65534`)
- A non-admin user is allowed to run one event using that Plugin.

**Action**

The non-admin user only clicks **Run** (or calls `run_event`). They do **not** send `debug_sudo`.

**Current behavior**

The manager currently copies the Plugin UID/GID only inside:

```js
if (isUnix) {
    if (plugin.uid) job.uid = plugin.uid;
    if (plugin.gid) job.gid = plugin.gid;
}
```

On a Windows manager, `isUnix` is false, so the job sent to the Linux worker has no UID or GID. The Linux worker then falls back to its own process identity:

```text
configured Plugin identity: nobody:nogroup (65534:65534)
job sent by Windows manager: no uid, no gid
Linux worker service identity: root:root (0:0)
actual spawned process: root:root (0:0)
```

So this is not “a non-admin can select sudo”. It is “a normal permitted run silently ignores the administrator's Plugin sandbox and uses the worker service account”. If the event is a shell-like job whose stored script or allowed argument is security-sensitive, the impact can be root execution. A fixed, non-user-controlled script is still executed under the wrong identity, but is not automatically arbitrary command execution.

### Concrete zero-value example

Numeric `uid: 0` / `gid: 0` is also lost by truthy checks such as `if (plugin.uid)` and `job.uid || process.getuid()`. For example, with a worker service UID/GID of `1000:1000`:

```text
Plugin configuration: uid=0, gid=0
expected spawn: 0:0
current spawn: 1000:1000
```

String `"0"` and numeric `0` can also behave differently across UI import/API paths. That is the reason for the zero-value parsing changes. This is primarily correctness/compatibility; the Windows-manager example above is the privilege-boundary issue.

### About `debug_sudo` and PR size

`debug_sudo` is a separate, admin-only feature. A non-admin should never be able to request it. The signed-dispatch code was added to transport that one-shot admin decision safely to the worker, but I agree that combining it with the Windows UID/GID bug and the zero-value cleanup made this much too large.

The GitHub diff also includes the stacked dependencies from #284, #257 and #286 because #287 is based on their commits while its PR base is still `main`. However, even the #287-only commit is still large: about 377 production additions plus 706 test additions. So your “super PR” criticism is fair.

I will not ask you to review #287 as a small fix in its current form, and I will stop opening new PRs while you digest the existing queue. If/when you want this revisited, the understandable split would be:

1. minimal platform-independent propagation of administrator-configured Plugin UID/GID, with the concrete Windows → Linux regression above;
2. numeric-zero UI/API compatibility;
3. admin-only `debug_sudo` transport hardening as a separate design review.

Thanks for the direct feedback — future descriptions will start with a concrete setup, action, expected result, and actual result before discussing implementation details.

mikeTWC1984:
OK, thanks. It made it clearer. 

I'll do more thorough review later, but at a glance

I see the problem with that now. But cant we just fix it by removing check for isUnix?
```js
if (isUnix) {
    if (plugin.uid) job.uid = plugin.uid;
    if (plugin.gid) job.gid = plugin.gid;
}
```

debug_sudo - yes it's only for admin. Run job api checks for admin privilege already before launching job. Why do we need extra hardening?

frankii91:
Yes. For the Windows manager -> Unix worker bug, the minimal fix is to move Plugin UID/GID copying outside the manager-side `isUnix` guard. The manager only serializes administrator-controlled Plugin configuration; the executing worker is the component that knows whether Unix UID/GID spawn options apply. Unix-specific `process.getuid()` handling must remain guarded because it is unavailable on Windows.

The signed-dispatch protocol is too broad for this fix. `launch_job` is already accepted only from the authenticated manager socket, so after the manager has made the authorization decision the worker currently treats that payload as trusted. If the authenticated cluster channel itself is considered mutable, then every execution field (`command`, arguments, environment, files, UID/GID, etc.) needs a separately reviewed transport-integrity design; signing only as part of this run-as fix would mix that larger boundary into a small platform bug. I therefore propose removing the HMAC negotiation from #287.

There is still one separate, concrete `debug_sudo` lifecycle gap in the manager API:

1. `validateOptionalParams` validates only keys present in its rules; it does not reject unknown keys.
2. `requireValidEventData` removes event-level `cwd`, `uid`, `gid`, and `env`, but currently not `debug_sudo`.
3. Consequently, `create_event` / `update_event` can persist `debug_sudo: 1` in the stored event, including for a non-admin account that has event create/edit privileges.
4. `run_event` removes a non-admin caller's request-level `params.debug_sudo`, but then merges the stored event with those params. The persisted marker therefore returns in the resulting job.
5. On a Unix manager, `launchJob` sees that marker and replaces the Plugin UID with `process.getuid()`.

So the existing `run_event` check protects the immediate request but not the complete create/update/run lifecycle. The small fix is to remove `debug_sudo` at the event persistence boundary and carry an authorized admin request only as a one-shot server-side launch decision. This does not require the HMAC / capability-negotiation implementation.

I can reduce #287 to the platform-independent Plugin UID/GID propagation fix based on current `main`, and leave numeric-zero handling and the `debug_sudo` lifecycle for separate review.

mikeTWC1984:
OK, can you submit Window/Unix and debug_sudo fixes in 2 different PRs,  so I can merge it sooner. Other stuff is still not very clear to me, it will likely take long time to review

frankii91:
Thanks — I split the two requested fixes into independent PRs based directly on current `main`:

- #292 — Windows manager → Unix worker Plugin `uid` / `gid` propagation
- #293 — one-shot, admin-only `debug_sudo` lifecycle

Both have green CI and apply cleanly in either order. Numeric-zero handling, the HMAC/capability design, and the separate #286 discussion were intentionally left out.

I am closing this stacked PR as superseded so the smaller fixes can be reviewed independently.

PR #289, opened 2026-08-28
Fix bundled defaults breaking legacy OAuth

### LINKED ISSUE #288: oauth with oidc to authelia broke on the latest docker image
On the latest docker image cronicle/edge i get

```
cronicle-edge | [1787917890.273][2026-08-28 13:51:30][cronicle-edge][7][User][debug][3][Requesting oauth token from:][https://auth/api/oidc/token] 
cronicle-edge | [1787917890.569][2026-08-28 13:51:30][cronicle-edge][7][User][error][3][oauth failed][fetch failed] 
cronicle-edge | [1787917890.569][2026-08-28 13:51:30][cronicle-edge][7][User][error][login][oath failed][]
```

I reverted to [1.14.5](https://hub.docker.com/layers/cronicle/edge/1.14.5/images/sha256-08071009ca2a9ca75224d117e75c8b5cde421acd5a573d69f27f50e4cff9e378) without changing any setting at all and it works again.

My .env:

```
CRONICLE_oauth__enabled=1

CRONICLE_oauth__client_id=Cronicle
CRONICLE_oauth__client_secret=secret

CRONICLE_oauth__redirect_uri=https://cronicle/api/user/callback

CRONICLE_oauth__authorize_url=https://auth/api/oidc/authorization
CRONICLE_oauth__token_url=https://auth/api/oidc/token
CRONICLE_oauth__user_url=https://auth/api/oidc/userinfo

CRONICLE_oauth__scope=openid profile email
CRONICLE_oauth__user_attribute=preferred_username
```

Not much more info I can give you. Authelia never logs any failure so I doubt it reached Authelia on the callback. Curl to the auth point works just fine in the container.

frankii91:
Fixes #288.

## Problem

The bundled `sample_conf/config.json` is installed as the active configuration for a fresh bundle and Docker image. Its non-empty `idp.example` values for `oauth.issuer` and `oauth.jwks_url` therefore acted as runtime defaults rather than documentation-only examples.

Legacy OAuth deployments such as the configuration reported in #288 override the OAuth endpoints through environment variables but do not set OIDC verification endpoints. Because `scope` includes `openid`, the provider returns an ID Token; the inherited example issuer/JWKS values then make Cronicle attempt OIDC validation against `idp.example`, and the callback fails before UserInfo.

This is not caused by an old Node runtime: the current Docker image uses Node v22.16.0.

## Change

- Leave optional OIDC and logout endpoint values empty in the installed defaults, preserving legacy OAuth-only behavior.
- Keep a complete machine-readable OIDC example under `sample_conf/examples/oauth.oidc.json` and document that examples are not runtime defaults.
- Log the underlying fetch cause when available instead of only the generic `fetch failed` wrapper.
- Add regression coverage for bundled defaults, the separate example, and OAuth error reporting.

Explicitly configured OIDC remains fail-closed and continues to validate signed ID Tokens through the configured issuer and JWKS endpoint.

## Validation

- Focused OIDC suite: 37/37 passed.
- Full `npm test`: 95/95 passed, 375 assertions.
- `node --check`, JSON parsing, and `git diff --check` passed.
- Built a fresh Docker image from this branch.
- Automated Docker flow with only the legacy OAuth environment variables completed the callback, returned `/#Home`, and issued a session cookie.
- Automated Docker flow with explicit issuer/JWKS configuration fetched JWKS, verified a signed RS256 ID Token, reached UserInfo, and issued a session cookie.
- Before the change, the same legacy OAuth Docker flow reproduced `oauth failed: fetch failed` with no session cookie.

The Docker build also reports the pre-existing `Could not resolve "mariadb/callback"` SQL bundling diagnostic while still exiting successfully. It occurs on the unmodified base and is intentionally outside this OAuth-focused change.


mikeTWC1984:
thanks for quick turnaround

PR #291, opened 2026-08-29
fix: the session_id is appended to url query paramet... in console.html

anupamme:
## Summary

`session_id` is currently included in the query string when loading live console output and job logs (`get_live_console`, `get_job_log` in `htdocs/custom/console.html`). Since this value authenticates the request, placing it in the URL can cause it to be recorded by access logs, proxy logs, browser history, or other URL-observing infrastructure.

This change moves the session ID to an HTTP request header (`X-Session-ID`) instead, while preserving existing API behavior.

## Why this works without any server-side change

`lib/api.js`'s `loadSession()` already reads the session ID with this precedence: cookie → `X-Session-ID` header → body param → query string (`args.cookies['session_id'] || args.request.headers['x-session-id'] || args.params.session_id || args.query.session_id`). Both `api_get_job_log` and `api_get_live_console` (`lib/api/job.js`) call `loadSession()` directly, so the header path was already supported — this PR only changes the client to use it.

## Verification

Manually verified end-to-end, per request:

1. Built and ran a local Cronicle instance (`node bin/build.js dev`, `node bin/storage-cli.js setup`, `node lib/main.js --foreground --manager`).
2. Logged in as `admin`, created a Shell Script event, and ran it.
3. Opened the Job Details live-log view while the job was in progress.
4. Clicked **View Full Log** — it opened `/console?id=<job_id>` in a new tab and rendered the job's log output correctly.
5. Confirmed via network inspection that both `get_live_console` and `get_job_log` requests returned `200 OK` with `X-Session-ID` sent as a header and **no `session_id` in the URL**.

## Changes
- `htdocs/custom/console.html`


mikeTWC1984:
agree, but can you test if it works. It may need to be lower case like x-session-id

anupamme:
Good catch to double check — turns out header names are case-insensitive in HTTP, and Node's `http` module always lowercases incoming header keys, so `fetch(url, {headers: {'X-Session-ID': ...}})` arrives server-side as `req.headers['x-session-id']`, which is exactly what `lib/api.js`'s `loadSession()` already checks. So the casing itself isn't an issue.

I added a regression test (`test/session-header.test.js`) that authenticates `get_job_log`/`get_live_console`'s shared session-loading path using only the header (no `session_id` in the URL) to confirm it actually works end-to-end, plus a check that the old query-string path still works for backward compatibility. `npm test` passes locally (100/100, 375 assertions). Also cleaned up the PR description to reflect this.

mikeTWC1984:
No need for extra test files. Build cronicle, run some test job, in live log view click on "view full log" it should open up console page, make sure you see the log

anupamme:
Done — dropped the automated test file and verified manually instead:

1. Built and ran a local instance (`node bin/build.js dev`, `node bin/storage-cli.js setup`, `node lib/main.js --foreground --manager`).
2. Logged in as `admin`, created a Shell Script event, and ran it.
3. From the Job Details live-log view (job still in progress), clicked **View Full Log** — it opened `/console?id=<job_id>` in a new tab and showed the log output correctly.
4. Checked the network tab: both `get_live_console` and `get_job_log` came back `200 OK`, authenticated via the `X-Session-ID` header, with no `session_id` anywhere in the URL.

Updated the PR description to reflect this. Let me know if you'd like me to test any other path (e.g. the download-log variant from the live view, or a completed job's "View Full Log" link).

PR #292, opened 2026-09-01
Fix Plugin run-as propagation from Windows managers

frankii91:
Split from #287 as requested.

## Problem

Setup: a Windows manager dispatches a Plugin with `uid` / `gid` configured to a Unix worker.

Expected: the worker receives the administrator-configured Plugin identity.

Actual: `launchJob()` copied these fields only when the manager itself was Unix, so the Windows manager dropped them and the Unix worker fell back to its service identity.

## Change

- Copy non-empty Plugin `uid` and `gid` into the job on every manager platform.
- Keep the existing `debug_sudo` service-UID lookup behind the Unix guard.
- Leave Unix-only child spawn handling on the executing worker unchanged.

This uses the existing job payload fields and requires no protocol change.

## Regression coverage

The test loads the job manager with a simulated `win32` platform, dispatches to a fake remote Unix worker, and verifies that Plugin `uid` / `gid` replace hostile event-level values in the transmitted job.

Validation:

- `npm test`: 97/97 tests, 443 assertions
- repository `Dockerfile`: image build completed successfully, including the SQL bundle

## Unchanged

- Numeric zero handling is not changed.
- `debug_sudo` lifecycle is not changed in this PR.
- Windows workers still do not receive Unix child-process options.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -292,11 +292,9 @@ module.exports = Class.create({
 
 					job.command = plugin.command;
 					if (plugin.cwd) job.cwd = plugin.cwd;
-					if (isUnix) {
-						if (plugin.uid) job.uid = plugin.uid;
-						if (job.debug_sudo) job.uid = process.getuid();
-						if (plugin.gid) job.gid = plugin.gid;
-					}
+					if (plugin.uid) job.uid = plugin.uid;
+					if (plugin.gid) job.gid = plugin.gid;
+					if (isUnix && job.debug_sudo) job.uid = process.getuid();
 
 					job.env = self.server.config.get('job_env') || {} 
 					let proto = 'http://'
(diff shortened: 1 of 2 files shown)


mikeTWC1984:
Looks good, but there was a conflict in test.js. Tried to resolve but test failing now (some missing comma) can you check

mikeTWC1984:
or maybe remove that test at all. I think it's a good idea to run some actual mixed setup for testing, I wouldn't be surprised if there are some other bug will pop up.

frankii91:
Thanks — checked. The merge conflict dropped the `},` that terminates the Windows-manager regression test. I restored only that separator in 0a6ab94.

`node --check lib/test.js` now passes, and the full `npm test` suite passes 101/101 tests with 465 assertions.

I kept the unit regression because it covers the manager-to-worker UID/GID dispatch payload. I agree that this is not a substitute for an actual mixed Windows-manager/Unix-worker setup; that end-to-end topology remains unverified here.

mikeTWC1984:
ok, good

PR #293, opened 2026-09-01
Keep debug sudo authorization one-shot

frankii91:
Split from #287 as requested.

## Problem

The Sudo checkbox is described as a manual, non-persistent debug option. However, `debug_sudo` was accepted as unknown event data, so it could be stored by create/update. `run_event` then checked only the caller-supplied field before merging the stored event, allowing a persisted marker to select the manager service UID later. A capacity queue could also retain the marker.

## Change

- Remove `debug_sudo` from all validated event data so create/update cannot persist it.
- Capture only explicit `true`, `1`, `"1"`, or `"true"` values from the manual run request.
- Authorize that one-shot choice only for an administrator and pass it separately from the event/job object.
- Materialize the manager service UID before the existing `launch_job` transport, then remove the marker before active state, hooks, queues, delays, or worker execution can retain it.
- Ignore stale top-level markers received by a newer worker.

The existing manager/worker protocol remains unchanged; no HMAC or capability layer is introduced.

## Regression coverage

Coverage verifies create/update persistence, admin manual execution, non-admin editor and event-token denial, remote dispatch identity, stale marker removal, and capacity-queue cleanup.

Validation:

- `npm test`: 100/100 tests, 456 assertions
- repository `Dockerfile`: image build completed successfully, including the SQL bundle

## Unchanged

- Plugin `uid` / `gid` behavior is not changed in this PR.
- Numeric zero handling is not changed.
- Existing manager-side `debug_sudo` resolution and mixed-version `launch_job` transport are preserved.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -264,6 +264,7 @@ module.exports = Class.create({
 		delete event.uid;
 		delete event.gid;
 		delete event.env;
+		delete event.debug_sudo;
 
 
 		// make sure title doesn't contain HTML metacharacters
--- a/lib/api/event.js
+++ b/lib/api/event.js
@@ -443,6 +443,10 @@ module.exports = Class.create({
 		if (params.id) criteria.id = params.id;
 		else if (params.title) criteria.title = params.title;
 		else return this.doError('event', "Failed to locate event: No criteria specified", callback);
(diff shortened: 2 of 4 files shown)


mikeTWC1984:
lgtm

PR #299, opened 2026-09-16
Fix OAuth before_login user context

### LINKED ISSUE #298: OAuth callback fires before_login before assigning args.user
## Summary

OAuth/OIDC login can fail during the `before_login` hook because the OAuth callback path fires the hook before populating `args.user`.

This appears to be a regression exposed by the September 7, 2026 sync to `0.9.131` (`6cce293f2732c5ee42fbdb645c87e49c3553ba2d`). That sync added schedule filtering in `beforeUserLogin()`, and the filtering path dereferences `args.user`. Password login already assigns `args.user` before `before_login`; the OAuth callback path assigns it only later, before `after_login`.

## Reproduction

1. Configure OAuth/OIDC login.
2. Log in through the OAuth callback path, e.g. `api_callback` in `lib/user.js`.
3. Use a build that includes the `0.9.131` sync where `beforeUserLogin()` filters schedule data by user privileges.
4. Observe login fail during the `before_login` hook.

## Stack trace essence

The failing path is effectively:

```text
api_callback OAuth login
  -> fireHook('before_login', args, ...)
  -> beforeUserLogin(args)
  -> filterScheduleForUser(..., args.user, ...)
  -> hasCategoryPrivilege(args.user, event.category)
  -> user.privileges.admin
  -> TypeError: Cannot read properties of undefined (reading 'privileges')
```

## Root cause

The OAuth callback path loads and validates the user, then sets only:

```js
params.username = userName;
```

and immediately fires:

```js
self.fireHook('before_login', args, function (err) {
```

At that moment, `args.user` is still undefined.

Later, after the session is created, the OAuth callback does:

```js
args.session = session;
args.user = user;
self.fireHook('after_login', args);
```

By contrast, the normal password login path sets `args.user = user` before proceeding through the login hooks, so the new schedule filtering code has the user object it expects.

## Proposed fix

Assign the authenticated OAuth user onto the hook args before firing `before_login`:

```js
params.username = userName;
args.user = user;
```

This keeps the OAuth callback hook contract aligned with password login and gives `beforeUserLogin()` the user object now required by schedule filtering.

Teagan42:
## Summary

- assign the validated OAuth user to `args.user` before firing the `before_login` hook
- add a regression test proving OAuth callback hooks receive the authenticated user

Fixes #298.

## Root cause

The OAuth callback path loaded and validated the user, then fired `before_login` before copying that user onto the hook args. The September 7, 2026 sync to `0.9.131` added schedule filtering in `beforeUserLogin()`, which now dereferences `args.user` via privilege checks. Password login already assigned `args.user` before its `before_login` hook; OAuth did not.

## Tests

- `npx pixl-unit test/oidc-logout.test.js`
- `npm test`


mikeTWC1984:
Good catch, thanks!

